PR #5, opened 2018-09-12
Enable thread pool for prediction

### LINKED ISSUE #3: Make predict use a thread pool
Predicting multiple directions is currently really slow because it uses a nested OpenMP for loop over i) direction; and ii) sources. When more than one direction is used, the second loop doesn't get multi-threaded anymore. In my case, I have 3 directions with ~100, 100 and 2000 sources, and because of this I have to wait forever. It's not hard to implement a parallel for loop with a thread pool, which would be the best approach with all combinations of directions/sources.

aroffringa:
These changes work and fix #3 . On Dawn nodes, DDE Cal was about two orders of magnitude faster because of this change :-/ !


aroffringa:
Needs a bit more work -- current solutions don't look good

PR #7, opened 2018-09-12
Use thread pool for predict

aroffringa:
Finished #3 . Solutions with predict & thread pool now look good as far as I can see. On Dawn I do as excepted see a CPU usage of ~4800% (all CPUs) during predict, with a 3-direction, many thousands of component model. Speed increase is also as expected a few tens of times (previous result of >100 seemed to be too good to be true -- and thus they were. Getting a recursively called thread pool right is actually not that easy!) Good improvement for doing DDE cals.

This thread pool could be a bit more efficient: right now it decides upon entry whether it should distribute the iterations or not (which it shouldn't when no threads are available) but when one direction finishes quickly and the nr of directions are larger than the nr of threads it could happen that not all cpu cores will be used.


DPPP/H5ParmPredict.h
```diff
@@ -85,6 +87,7 @@ namespace DP3 {
       std::vector<std::string> itsDirections;
 
       NSTimer          itsTimer;
+			ThreadPool itsThreadPool;
     };
```
tammojan: Tabs should be spaces in this file

DPPP/Predict.cc
```diff
@@ -310,38 +313,50 @@ namespace DP3 {
...
-        // Keep on predicting, only apply beam when an entire patch is done
-        if (itsApplyBeam && curPatch!=itsSourceList[i].second && curPatch!=0) {
-          addBeamToData (curPatch, time, refdir, tiledir, thread, nSamples,
-                         itsModelVisPatch[thread].data());
-        }
-        simulator.simulate(itsSourceList[i].first);
-        curPatch=itsSourceList[i].second;
-      }
-
+			std::unique_ptr<ThreadPool> localThreadPool;
```
tammojan: This files uses spaces for indentation...

DPPP_DDECal/DDECal.cc
```diff
@@ -617,12 +618,16 @@ namespace DP3 {
                                 itsModelData[itsStepInSolInt]);
         itsModelDataPtrs[itsStepInSolInt][0] = itsModelData[itsStepInSolInt].data();
       } else {
-#pragma omp parallel for schedule(dynamic) if(itsPredictSteps.size()>1)
-        for (size_t dir=0; dir<itsPredictSteps.size(); ++dir) {
+//#pragma omp parallel for schedule(dynamic) if(itsPredictSteps.size()>1)
```
tammojan: spaces here please..

tammojan:
Since you're asking for a review:
 * Could you squash the commits into one
 * Could you follow standard [commit message guidelines](https://chris.beams.io/posts/git-commit/), i.e. "Use threadpool for Predict"

aroffringa:
I've fixed the tabs. Isn't squashing easiest when merging? (Or this is a suggestion I squash + merge? -- I thought that was the reviewer's role :) )

tammojan:
You're right, I can squash them while merging. We'll debate some other time whose role it is ;)

PR #13, opened 2018-09-15
Disable simulator parallelization

### LINKED ISSUE #11: Disable OMP parallelization in Simulator
Now that the threadpool is used, it turns out there was an extra `#pragma omp parallel` for in the Simulator. Since this used to be a nested omp loop, it had no effect, but now that the threadpool takes care of the parallelization, omp does parallelize this loop and starts a new group of threads in each threaad, giving rise to 48^2 threads on my 48 core machine.

aroffringa:
Fixes #11, which was related to #3 . This pool contains some left-over cleaning of OMP statements that caused slower performance, and implements a new ThreadPool. This ThreadPool has better performance and keeps all new loops running, causing it to occupy the full CPU as intended.

### DIFF
--- a/Common/ThreadPool.h
+++ b/Common/ThreadPool.h
@@ -63,7 +63,7 @@ public:
 		++_priority;
 		lock.unlock();
 		
-	  size_t progress = end-start;
+		size_t progress = end-start;
 		
 		std::thread localThread(&ThreadPool::threadSpecificPriorityFunc, this, 0, thisPriority, &progress);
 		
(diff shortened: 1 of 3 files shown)


Common/ThreadPool.h
```diff
@@ -54,54 +58,107 @@ class ThreadPool
...
 	{
-	  size_t progress = 0, n = end-start;
-		
 		std::unique_lock<std::mutex> lock(_mutex);
-		if(_freeThreads == 0)
+		size_t thisPriority = _priority;
+		++_priority;
+		lock.unlock();
+		
+	  size_t progress = end-start;
```
tammojan: Please use consistent indentation.

PR #29, opened 2018-09-25
Make DPPP compile without LOFAR beam

### LINKED ISSUE #28: Make DPPP compile without LOFAR beam
I'm actually trying DPPP on MWA data, but it is a bit of an issue to compile the LOFAR repository on MWA nodes

aroffringa:
Fixes #28. Steps to correct or apply the beam are put between #ifdefs and not compiled if no beam is found. The demixen is not compiled either when no beam is available, as it does not make sense to use demixen for anything other than LOFAR (or just anything?)

Because not any LOFAR files are any longer included when building without beam, in various places new includes or similarly slight changes were necessary.


tammojan:
Great, thanks!

PR #30, opened 2018-09-25
Fix inconsistent use of tabs

tammojan:
(This only removes tabs from files where all the rest of the file is space indented. Please keep code style consistent within files.)


aroffringa:
Thanks for fixing my untidiness. I've now setup my editor properly so I'll keep it clean :)

Issue #31, opened 2018-09-26
Setting startchan when updating column writes wrong channels


aroffringa:
MWA datasets are large and need to be split up by channel and processed by running DPPP several times. This should in principle be possible with the `msin.nchan` and `msin.startchan` parameters, however due to a bug MSUpdater does not actually use offset the channels at the proper index. This is because the MSReader never sets `startchan` in the MSInfo object, only `nchan`.

PR #32, opened 2018-09-26
Write to correct channel indices when processing only part of the channels

aroffringa:
Solves #31 . Tested on an MWA set; issue solved.

### DIFF
--- a/DPPP/DPInfo.cc
+++ b/DPPP/DPInfo.cc
@@ -61,7 +61,7 @@ namespace DP3 {
                        const string& msName, const string& antennaSet)
     {
       itsNCorr        = ncorr;
-      itsStartChan = startChan;
+      itsStartChan    = startChan;
       itsNChan        = nchan;
       itsOrigNChan    = nchan;
       itsNTime        = ntime;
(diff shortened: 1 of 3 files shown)


DPPP/DPInfo.cc
```diff
@@ -56,11 +56,12 @@ namespace DP3 {
         itsPhaseCenterIsOriginal (true)
     {}
 
-    void DPInfo::init (uint ncorr, uint nchan,
+    void DPInfo::init (uint ncorr, uint startChan, uint nchan,
                        uint ntime, double startTime, double timeInterval,
                        const string& msName, const string& antennaSet)
     {
       itsNCorr        = ncorr;
+      itsStartChan = startChan;
```
tammojan: Please use consistent indentation in files.

aroffringa: Are you referring to the fact that the equal signs don't align? I'll fix it, but I find this kind of over-spacing really annoying and a waste of time, it just brings work to keep it like this and is really not more readable. In fact, sometimes the equal sign in DP3 is so far from the variable that you have to place your finger on the screen to see which is which. Proper tooling will already highlight and colour the code, which makes it as readable as possible.

PR #35, opened 2018-09-27
Add old casacore include dir

tammojan:
This is to work around the fact that StationResponse.h until very recently used the old casacore include path (without the casacore prefix). The error thrown during build was:
```
In file included from /home/installdir/lofar/include/StationResponse/AntennaField.h:36:0,
                 from /home/installdir/lofar/include/StationResponse/Station.h:32,
                 from /home/installdir/DP3/DPPP/DPInput.h:36,
                 from /home/installdir/DP3/DPPP/MSReader.h:30,
                 from /home/installdir/DP3/DPPP/DPRun.h:31,
                 from /home/installdir/DP3/DPPP/DPRun.cc:24:
/home/installdir/lofar/include/StationResponse/ITRFDirection.h:32:41: fatal error: measures/Measures/MeasFrame.h: No such file or directory
compilation terminated.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,6 +20,9 @@ include_directories(${HDF5_INCLUDE_DIRS})
 
 find_package(Casacore COMPONENTS casa ms tables REQUIRED)
 include_directories(${CASACORE_INCLUDE_DIR})
+# Add old casacore include directory because LOFAR beam library did
+# not use the casacore/ prefix in the includes until 3.2 (Sep '18)
+include_directories(${CASACORE_INCLUDE_DIR}/casacore)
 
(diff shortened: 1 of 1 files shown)


aroffringa:
Personally I'd rather depend on the newest LOFAR/Casacore repos than to add compatibility fixes. For those not using the newest LOFAR, they don't need the newest DP3 either...

tammojan:
In principle I agree, but StationResponse has been stable for years except for this include dir. It would be silly to require people to upgrade LOFAR just because of this. Specially because the fix was only present in the very lastest LOFAR release.

aroffringa:
Ok, I agree. It would be good to put a brief comment above the line to explain why it is there.

aroffringa:
Nice :)

Issue #27, opened 2018-09-25
Segfault in DDECal solve


darafferty:
I am getting a segfault when doing a DDECal solve with the following command:

`DPPP msin.datacolumn=DATA solve.propagatesolutions=True solve.sourcedb=calibration_skymodel.make_sourcedb numthreads=6 msin.starttime="09Dec2016/00:09:21.076" solve.solint=1 solve.mode=tec solve.maxiter=50 solve.type=ddecal steps=[solve] solve.tolerance=0.0001 msin=chunk2.ms solve.nchan=1 solve.stepsize=0.2 solve.smoothnessconstraint=3e5 msin.ntimes=0 solve.approximatetec=True solve.maxapproxiter=25 solve.h5parm=fast_phase_35.h5parm solve.uvlambdamin=80.0 msout=.`

Tammo Jan suggested using gdb to debug it, and as far as I can tell, it looks like it has something to do with a NULL pointer dereference in `DP3::DPPP::Predict::process()`. Changing line 364 of https://github.com/lofar-astron/DP3/blob/master/DPPP/Predict.cc to `for (size_t thread=0; thread!=pool->NThreads(); ++thread) {` fixed the segfault, but I don't really know if that's the correct fix (my knowledge of c++ is pretty limited!). I can provide the MS file, sky model, etc. if needed.

aroffringa:
Thanks! Hmm your fix _should_ not do anything, as pool->NThreads() should be initialized to OpenMP::maxThreads() in the more recent version of DPPP. This was not the case before #6 was fixed though, so could it be that you were/are using a version from before that fix? Otherwise I've introduced apparently another bug somewhere.

darafferty:
It was with the latest master -- I put the files on ftp.hs.uni-hamburg.de/pub/outgoing/rafferty so that you can try (sorry, the MS is rather large).

aroffringa:
Hi @darafferty , I downloaded your data and can reproduce it. Your fix seems okay. However, I don't understand why it is necessary and would like to: @tammojan it looks like `OpenMP::maxThreads()` has returned a higher value in DDECal (where the threadpool is created) than it does return later on in Predict. It crashes when ThreadPool launches the seventh thread, while openMP at that point returns 6 (as set by the keywords), but the threadpool was earlier initialized from `OpenMP::maxThreads()`, which at that point probably returned the full nr of cores (48 in my case). Is this possible?

tammojan:
[...]

PR #36, opened 2018-09-27
#27 Avoiding combination of OpenMP and normal threads

aroffringa:
This pull is rather large and affects several steps, including AOFlagger, H5ParmPredict, Predict and DDECal. It is not tested yet.


aroffringa:
I've tested it somewhat, and as far as I can tell everything works, so I'm merging it.

PR #50, opened 2018-10-03
Aoflagger does not load because of missing libraries

aroffringa:


aroffringa:
@tammojan would you like to fix Travis-CI ? Seems it misses some libs.

tammojan:
Was boost-signals really necessary?

aroffringa:
I use it in aoflagger... don't think it is trivial to remove

tammojan:
Ok, I'll fix travis.

PR #52, opened 2018-10-03
Linking aoflagger, ddecal and interpolate directly into DPPP

aroffringa:


aroffringa:
@tammojan is this ok with you?

PR #58, opened 2018-10-11
Fixes bug reported by Aleksandar due to nthreads unset

aroffringa:
This bug was introduced by #27 (avoiding OMP). Tested on Dawn.

### DIFF
--- a/DPPP/H5ParmPredict.cc
+++ b/DPPP/H5ParmPredict.cc
@@ -25,7 +25,6 @@
 
 #include "Exceptions.h"
 
-#include "../Common/OpenMP.h"
 #include "../Common/ParameterSet.h"
 #include "../Common/StreamUtil.h"
 #include "../Common/StringUtil.h"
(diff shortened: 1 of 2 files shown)


DPPP/H5ParmPredict.cc
```diff
@@ -110,19 +109,18 @@ namespace DP3 {
...
       info().setWriteData();
 
-      vector<Predict::ShPtr>::iterator predictstep;
-      for (predictstep=itsPredictSteps.begin();
-           predictstep!=itsPredictSteps.end();
-           predictstep++) {
-        (*predictstep)->updateInfo(infoIn);
+      for(Predict::ShPtr& predictstep : itsPredictSteps)
+      {
+        predictstep->setNThreads(NThreads());
```
aroffringa: This is the actual fix :)

tammojan: Would it be more useful to embed nthreads in DPInfo? That way you cannot forget to set it (it's set to a default in DPRun I think).

aroffringa: I considered it, but I thought there was some value in having a separation of things which come out of the MS and program options -- e.g. the parset is also not part of DPInfo. Separately setting nthreads was a bit more of a hassle than I thought though, so whether it was the right choice, hmm. We can reconsider if we change things related to it.

PR #64, opened 2018-10-14
Fix missing threads lib and incorrect Boost-Python detection

### LINKED ISSUE #56: Python install script doesn't compile on Galaxy
On Galaxy (one of the cluster of the Pawsey computer centre), there's no boost python. This is fine, but it seems that the python install scripts don't handle a missing boost python properly:
 
```
CMake Error at CMake/PythonInstall.cmake:120 (install):
  install FILES given unknown argument
  "/group/mwa/software/aoflagger/v2.12.1/galaxy/lib/python2.7/site-packages/lofar/dppp/".
Call Stack (most recent call first):
  DPPP/CMakeLists.txt:27 (python_install)


WARNING, Boost-Python not found; PythonDPPP will not be built.
-- Configuring incomplete, errors occurred!

```

Just removing the python install things from line DPPP/CMakeLists.txt:27 fixes this, but obviously doesn't install pythondppp anymore.

aroffringa:
Fixes #56 . Tested on Galaxy.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,6 @@ cmake_minimum_required(VERSION 3.9)
 
 project(DP3)
 
-# Casacore has a separate CMake file in this directory
 set(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/CMake)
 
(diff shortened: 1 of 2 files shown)


aroffringa:
I think what happened was that the python install script can't handle an empty file list. By fixing the check on boost python, which this PR solves, the file list is no longer empty.

Issue #73, opened 2018-11-05
Predict/ThreadPool problem: std exception detected: vector::reserve


darafferty:
At the start of a GainCal step (from Prefactor), I get the error "std exception detected: vector::reserve". gdb gives the following info:

```
Processing 1798 time slots ...

0%[New Thread 0x15553c4da700 (LWP 31675)]
[New Thread 0x15553c2d9700 (LWP 31676)]
[New Thread 0x15553c0d8700 (LWP 31677)]
[New Thread 0x15553bed7700 (LWP 31678)]
[New Thread 0x15553bcd6700 (LWP 31679)]
[New Thread 0x15553bad5700 (LWP 31680)]

Thread 1 "DPPP" hit Catchpoint 1 (exception thrown), 0x000015554f3af8bd in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
(gdb) bt
#0  0x000015554f3af8bd in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#1  0x000015554f3d826f in std::__throw_length_error(char const*) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#2  0x00000000004deba8 in std::vector<std::thread, std::allocator<std::thread> >::reserve(unsigned long) ()
#3  0x00000000005864d8 in DP3::ThreadPool::ThreadPool(unsigned long) ()
#4  0x00000000005846ea in DP3::DPPP::Predict::process(DP3::DPPP::DPBuffer const&) ()
#5  0x000000000055b3a6 in DP3::DPPP::GainCal::process(DP3::DPPP::DPBuffer const&) ()
#6  0x0000000000512e92 in DP3::DPPP::Filter::process(DP3::DPPP::DPBuffer const&) ()
#7  0x000000000049b369 in DP3::DPPP::MSReader::process(DP3::DPPP::DPBuffer const&) ()
#8  0x000000000045782f in DP3::DPPP::DPRun::execute(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, int, char**) ()
#9  0x0000000000454c1e in main ()
```
The command I used was:

`
[...]

PR #74, opened 2018-11-07
Fix two nthread related crashes reported by David

aroffringa:
This fixes the vector reserve exception #73 and the 2nd segfault reported in that bug report, both crashes occuring in the model predictor when using gaincal.

NThreads is now moved to the DPInfo object, as earlier suggested also by @tammojan , which avoids all the manual setting.

### DIFF
--- a/Common/ThreadPool.h
+++ b/Common/ThreadPool.h
@@ -40,7 +40,8 @@ public:
 		_priority(0)
 	{
 		size_t nThreads = NCPUS();
-		
+		if(nThreads == 0)
+			nThreads = 1;
 		// We reserve one thread less, because we always want a new For loop
 		// to be able to add a new thread (with index 0).
 		_threads.reserve(nThreads-1);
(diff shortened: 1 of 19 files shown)


Common/ThreadPool.h
```diff
@@ -40,7 +40,8 @@ class ThreadPool
 		_priority(0)
 	{
 		size_t nThreads = NCPUS();
-		
+		if(nThreads == 0)
+			nThreads = 1;
```
tammojan: Does this mean the default number of threads becomes 1? I think the number of CPUs was the previous default (and is probably a more sensible default).

aroffringa: Not quite: it only becomes one when NCPUS() returns zero (this can maybe happen when all CPUs are scheduled for another task, since NCPUS() takes affinity into account). However, this overload of the constructor is never used in DP3, only the one with explicit nthread count is used. I've submitted some additional comments.

aroffringa: Oh maybe you've missed the `if` statement above the nThreads = 1...

Issue #60, opened 2018-10-12
Add keywords that describe how beam is applied


aroffringa:
This is really necessary to avoid people apply incorrect beams (i.e. fuil instead of differential), which I unfortunately see happening a lot!

tammojan:
It would be good to register here also the name ('hamaker-arts') and version of the beam model applied.

aroffringa:
I think that's less crucial. When we have multiple versions of the beam, we could always add those fields without breaking backward compatibility (everything that has beam keywords is assumed to have the version of the beam that we have now). I think we're a bit overdesigning when we're adding those now already...

aroffringa:
Fixed in #77 .

PR #77, opened 2018-11-09
Store how beam is applied

aroffringa:
Adds the columns LOFAR_APPLIED_BEAM_DIR and LOFAR_APPLIED_BEAM_MODE to the field table. Rerunning applybeam on a measurement set will now result in an error (if necessary, later it could be made possible to undo the beam-application and remove the keywords, or even to change it from one direction to another). Ticket #60 .


tammojan:
Why do you store the info in the FIELD table? For data sets with multiple data columns (which is arguably against the MS standard), I suppose it could also go as a column keyword of the data column. But also happy to merge as is.

aroffringa:
I agree with you, that would be better. I chose the field column because TILE_DIR is also there (and others), and this fell in the same category, but you're right that it would be handy to actually be able to have different beam keywords for different columns (very likely to occur...) I'll change the merge request.

aroffringa:
I've now implemented it with data column keywords, which is indeed better! (and actually less code :D). Below is how it looks like after applying beam to `DATA` and copying that column to `CORRECTED_DATA`. Ready to merge, but be sure to check whether you agree with this because it is after all a change to the MS format, and we should get that right first time.

image

adrabent:
What happens if the CORRECTED_DATA is beam-corrected and I would overwrite the CORRECTED_DATA column with a non-beam corrected DATA column. Will this information be removed then?

aroffringa:
Good question. It's indeed supposed to change the information to reflect the new state, but that's one case I haven't tested. (Note that in this case (and it works correctly) it wouldn't actually remove the keywords, but set the keywords to reflect 'no beam applied').

PR #88, opened 2018-11-23
Interpolate applycal

tammojan:
This adds optional bilinear interpolation for ApplyCal, on frequency and time. Default is to keep doing nearest neighbor interpolation.

Linear interpolation could be very useful if some data is omitted during calibration.

The interpolation is on the real values in the H5Parm, so on either the amplitudes (where it makes sense) or on the phases (where linear interpolation doesn't necessarily makes sense).

I'm not sure if interpolating the weights is the right thing to do.

### DIFF
--- a/DPPP/GridInterpolate.cc
+++ b/DPPP/GridInterpolate.cc
@@ -96,7 +96,50 @@ namespace DP3 {
         }
       }
     } else {
-      throw std::logic_error("Not implemented");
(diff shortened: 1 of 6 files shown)


tammojan:
(I'm not particularly proud of this code, but it should do the trick. It's probably not too efficient but this is not a critical piece of code.)

Issue #91, opened 2018-11-26
Specify in output  where metadata does not match 


maaijke:
I have 2 AARTFAAC MS with slightly diffrerent times (subsecond difference). I tried to copy the data for every column containing time information in the MS, however I still get the exception that the metadata differs.

In DPPP MultiMSReader.cc we hit this exception:

          if (!(near(itsStartTime, rdinfo.startTime())  &&
                     near(itsLastTime, itsReaders[i]->lastTime())  &&
                     near(itsTimeInterval, rdinfo.timeInterval())  &&
                     itsNrCorr == rdinfo.ncorr()  &&
                     itsNrBl   == rdinfo.nbaselines()  &&
                     itsFullResNChanAvg == itsReaders[i]->nchanAvgFullRes()  &&
                     itsFullResNTimeAvg == itsReaders[i]->ntimeAvgFullRes()  &&
                     getInfo().antennaSet() == rdinfo.antennaSet()  &&
                     allEQ (getInfo().getAnt1(), rdinfo.getAnt1())  &&
                     allEQ (getInfo().getAnt2(), rdinfo.getAnt2())))
            throw Exception(
                     "Meta data of MS " + itsMSNames[i]
                     + " differs from " + itsMSNames[itsFirst]);

 

Could it be separated such that the exact mismatch is clear?

aroffringa:
@maaijke : I've updated the DP3 version on Dawn at:

`/home/users/offringa/Software/DP3/build/DPPP/DPPP`

So you could use that version to test what's the problem in your MSes.

PR #92, opened 2018-11-26
Extend error msg when multims metadata differs

aroffringa:
Implements #91.


tammojan:
Thanks! As a tip: please instead of 'Implements', use 'closes', 'resolves' or 'fixes', so github will [automatically](https://help.github.com/articles/closing-issues-using-keywords/) close the issue for you.

aroffringa:
Maaijke: I've updated the DP3 version on Dawn at:

`/home/users/offringa/Software/DP3/build/DPPP/DPPP`

So you could use that version to test what's the problem in your MSes.

aroffringa:
@tammojan Thanks, didn't know that.

Issue #81, opened 2018-11-15
in low S/N regime some stations are set to 0 and some other to high values


revoltek:
when I solve diag+rot or diag in regions with noisy data **some** stations are set at 0, **some others** at very high values (the effect varies with different weights). Note that in many cases the first ~half of the stations are set to high values and the second half to 0.

See: https://www.dropbox.com/s/5qlfqi5bg9ol5pz/amp_polXX.png?dl=0

aroffringa:
Could you provide a testset for this? I'm not sure I can look at this any time soon.

aroffringa:
(reposting image here to keep it together with the bugreport -- @revoltek you can paste images right into the ticketbox :-) )
![image](https://user-images.githubusercontent.com/11645549/48670423-f1e6a580-eb17-11e8-9bbc-9fd695009984.png)

aroffringa:
@revoltek Can you post the parset that gives trouble as well?

revoltek:
I've tested it with:

msin =
msin.datacolumn = DATA
msout = .
msout.datacolumn = CORRECTED_DATA

steps = [sol]

sol.type = ddecal
sol.uvlambdamin = 300
sol.mode = diagonal
sol.maxiter = 500
sol.tolerance = 1e-3
sol.nchan = 1
sol.solint = 1
sol.propagatesolutions = True
sol.usemodelcolumn = True

revoltek:
NOTE: I don't see the problem with the DIE solver (same parameters)

revoltek:
This is another example with and without propagate solution. Note that the problem is there (bottom left corner) even with the porp-sol=False, it is just much less problematic as it doesn't spread all over the osbervation.

![amp_polxx](https://user-images.githubusercontent.com/4458741/48936318-da8a2c80-ef0a-11e8-9c84-e806bea3c97c.png)
![amp_polxx](https://user-images.githubusercontent.com/4458741/48936325-e0800d80-ef0a-11e8-8680-44a17cf7b38e.png)

revoltek:
I have at least one example where the DIE solver has the problem enhanced by the prop sol. Conclusion: both algorithm have the problem, propagate solutions makes it worse in both cases. Somewhat the magnitude of the effect is different if using DDE or DIE solver.

aroffringa:
I have analyzed it somewhat. It has to do with the following things:
[...]

PR #93, opened 2018-11-26
Replace missing solutions by average gain

aroffringa:
During solution propagation, instead of replacing missing (nan) solutions by unity, this commit will replace missing solutions by the average absolute gain. I decided to use average absolute instead of normal (complex) average, because otherwise this might result in near-zero result when the phase changes a lot over stations. This is related to #81.

I've tested it on one of Francesco's sets, showing that i) it is correctly implemented and ii) it solves #81 in that particular case, but it's not yet clear whether #81 is solved in its entirely.

### DIFF
--- a/DDECal/Matrix2x2.h
+++ b/DDECal/Matrix2x2.h
@@ -226,14 +226,15 @@ public:
 		double trHalf = tr*0.5;
 		e1 = trHalf + term;
 		e2 = trHalf - term;
-		if(matrix[2] != 0.0)
+		double limit = std::min(std::fabs(e1), std::fabs(e2)) * 1e-6;
+		if(std::fabs(matrix[2]) > limit)
 		{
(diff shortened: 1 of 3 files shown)


rvweeren:
Is there still a way to recognize that these came from "missing" solutions (via the weights?). Otherwise that creates problem for losoto (and our other scripts that work on H5 files) to search for outliers, etc.

aroffringa:
@rvweeren missing solutions are still given no weight. The problem was that the initial value in the interval *after* the missing solutions gave rise to divergence. Only the initial values are changed: the final solution is then either a converged solution (which is like any other converged solution) or, if it doesn't converge / have no unflagged visibilties, it is given no weight.

So, no you can't see that the initial value given to the solver was different, but there should be no reason for Losoto to know that.

rvweeren:
@aroffringa thanks for clarifying, all ok then.

PR #98, opened 2019-01-11
Add minvisratio feature

### LINKED ISSUE #83: Minimum number of visibilities in solve
In certain cases there are several stations flagged and the solver tries anyway to find solutions, obtaining noisy results. Can you add a "minimum number of baselines" option that if not satisfied flag the solution? Alternatively the same result can be achieved with a flagger step. Not sure what is the best approach.

aroffringa:
This allows skipping the solve for solution intervals with fewer unflagged visibilities. This fixes #83. So far I've done some quick testing with minvisratio=0.6, which seemed to work as usual, and minvisratio=1, which indeed flagged almost everything, so it seems to work on first order. Awaiting further validation by @revoltek.

### DIFF
--- a/DDECal/DDECal.cc
+++ b/DDECal/DDECal.cc
@@ -75,1006 +75,1056 @@ using namespace casacore;
 using namespace DP3::BBS;
 
 namespace DP3 {
-  namespace DPPP {
+namespace DPPP {
 
-    DDECal::DDECal (DPInput* input,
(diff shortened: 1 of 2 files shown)


revoltek:
Should I see in the log that the minvisratio is active?

aroffringa:
Oh, no it is not shown yet. Will do that...

PR #107, opened 2019-01-31
Reuse threads in GainCal's parallel for

aroffringa:
I noticed that GainCal is slower than it used to be. I'm not sure why, but a lot of threads are restarted
during the parallel for structures. This commit keeps the threads running between iterations.

This required a barrier. Unfortunately, the barrier in boost is buggy and the barrier in std requires
c++20 which we probably want to avoid, so I implemented a very simple one myself.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@ project(DP3)
 
 set(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/CMake)
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -std=c++11")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -std=c++11 -O3 -march=native")
 
 find_package(HDF5 COMPONENTS C CXX REQUIRED)
 add_definitions(${HDF5_DEFINITIONS})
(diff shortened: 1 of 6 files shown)


aroffringa:
Unfortunately I don't see a difference in speed from this commit. Not a bad change anyway, but have to do some more digging to find the speed bottleneck.

aroffringa:
Hmm something is not working yet, please don't merge yet.

aroffringa:
@revoltek the h5parmpredict task is pretty slow, although I wouldn't know why it would have become slower. Its slowness was the reason to implement subtraction straight in DDECal, which is faster.

revoltek:
> @revoltek the h5parmpredict task is pretty slow, although I wouldn't know why it would have become slower. Its slowness was the reason to implement subtraction straight in DDECal, which is faster.

yes, that can work in some cases, but we are now calibrating on a set of bright sources (after removing the rest of the field with DIE calibration) and we need to subtract the entire facet once the DDE calibration is done

aroffringa:
@revoltek I think h5parmpredict could be made as fast as the DDECal predict, but I think this is quite a bit of work. A shortcut is to enable a "DDECal without calibration" feature, that just predicts and subtracts/add/sets the visibilities to that value. You can open a ticket for this if this is useful.

Issue #99, opened 2019-01-11
Solver drifts away from chi square minimum


revoltek:
@darafferty and I noticed that, with a higher tolerance, solutions show less outliers (the opposite of the expectation) - this has been tested on DDE solver diag+rot, 1 direction. In the image attached left is with tolerance 0.1, right is with tolerance 0.001. Also the problem outlined in #81 seems less prominent in this case. It seems that the solver drift away from the correct minimum if it has enough iterations to do it.

![screenshot_20190111_133230](https://user-images.githubusercontent.com/4458741/51033941-5157b380-15a5-11e9-99d3-58db576e0c98.png)

aroffringa:
Interesting. I don't think this necessarily implies the solver drifts away from the least-squares solution. What you find is that the solutions are noisier, but solutions closer to the least-squares solutions are probably noisier than earlier solutions. It could very well be (and I think that is the case) that the squared-error of the solutions with lower tolerance is in fact smaller. It could also be that there are visibilities that don't match the model well, but which are benign in the first iterations, but start to have more effect later in the calibration.

Do the solutions with lower tolerance also show worse imaging tolerance? Apart from the parts where it diverged because of too little information, I expect not much difference between calibration. I think would be that, which this calibration strategy, iterating for more than tolerance=0.1 iterations is not necessary.

There is of course a strong frequency relation which could be used to reduce the noise / divergence.

I assumer the lower tolerance uses a lot more iterations? What is the difference in the number of iterations that are used?

revoltek:
@aroffringa @tammojan this seems to be a good workaround to preserve the propagatesolution=T, but it's still not clear to us why in those timestamps the solver has this strange behavior of pushing half of the antennas up and half down. If it will turn out that it is caused by "bad data", a possibility is to check if David's criterion is met, then flag the datapoint.

Maybe @maaijke is interested in this thread as she developed part of the diag+rotational solver.

aroffringa:
@darafferty Looks indeed very good, and indeed such a constraint makes sense. Did you implement that constraint within DP3 DDECal? I'm happy to merge it if so...

In your solutions, weren't those 'blocks' of bad solutions caused by a flagged/bad timestep already solved by commit #93 (which was an effort to fix #81 ) ?

darafferty:
@aroffringa: Yes, I implemented it within DDECal -- I will submit a pull request for it (though it hasn't had extensive testing yet...).

[...]

PR #111, opened 2019-02-04
Add additional options helpful for LBA calibration

darafferty:
I added a few things to DDECal to address some issues @revoltek found with calibration of LBA data (issues #81 and #99):

- add option to restrict propagation to converged solutions only ("propagateconvergedonly")
- add option to flag unconverged solutions ("flagunconverged")
- add an amplitude-divergence check to the rotation+diagonal constraint, as station amplitudes occasionally diverge (some with very high values, some with very low ones) in this mode. The check ensures amplitudes drift by no more than a factor of 5 from the mean over all stations
- add option to flag solutions if this amplitude divergence happens and the solver does not converge ("flagunconverged" + "flagdivergedonly"). All stations will be flagged if any one station is detected to diverge


revoltek:
I checked these options they seem to work as expected for me.

aroffringa:
I went through the code changes and it looks good to me. Thanks a lot @darafferty , seems this is very useful!

aroffringa:
@darafferty Btw, have you updated the DPPP doc at https://www.astron.nl/lofarwiki/doku.php?id=public:user_software:documentation:ndppp#ddecal with the new options ?

darafferty:
Ah, no -- thanks for reminding me. I'll do it now.

PR #118, opened 2019-02-11
Move showsourcedb from LOFAR repos to DP3

### LINKED ISSUE #117: Showsourcedb not moved to DP3
Showsourcedb is a simple parser for sourcedb's, can be useful to check if you made a mistake with makesourcedb. Should be easy to move from LOFAR as well.

aroffringa:
Fixes #117.


tammojan:
Thanks André!

PR #122, opened 2019-02-15
Revert "Movingphasedir"

aroffringa:
Reverts lofar-astron/DP3#121


aroffringa:
Hmm trying to change one thing after merging, but I've no idea how to do this with Github :-/ Guess I need to create a new pull separate request for that.

PR #127, opened 2019-03-07
Add direction table in rotation/rot+diagonal solve

### LINKED ISSUE #97: additional direction table in DDEcal for h5parms
Hi,

I encountered some kind of "inconsistency" in DPPP when handling h5parm in `DDEcal`.
If I use DDEcal in the mode `rotation+diagonal` I just get a solution table, like `phase000` with 120 times, 60 ants, 1484 freqs, 2 pol.
But If I just change to the mode `diagonal` in `DDEcal` I get a solution table, like `phase000` with 120 times, 60ants , 1484, 1 dir, 2 pol. Where does this direction table come from and is it really needed if solving only for one direction. And why does it not appear if one solves for common rotation angle?

Alex

aroffringa:
This possibly fixes #97, but it is untested.

### DIFF
--- a/DDECal/RotationAndDiagonalConstraint.cc
+++ b/DDECal/RotationAndDiagonalConstraint.cc
@@ -13,24 +13,26 @@ void RotationAndDiagonalConstraint::InitializeDimensions(size_t nAntennas,
                                                          size_t nChannelBlocks) {
   Constraint::InitializeDimensions(nAntennas, nDirections, nChannelBlocks);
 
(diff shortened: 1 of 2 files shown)


tammojan:
Looks ok to me!

Indeed I implemented this for one direction only; I'm not sure it makes sense physically to do this kind of calibration for multiple directions: the rotation usually is to correct for slight errors in the element rotation in the stations. Those do not differ per direction.

PR #130, opened 2019-03-08
When smoothing, set missing solutions to NaN

### LINKED ISSUE #128: smooth constraint + flag
It looks like that when the smooth constraint is active no all flags are properly propagated into the h5parm (e.g. fully flagged antennas)

aroffringa:
This partly fixes #128, but the weights themselves are not smoothed, so missing channels might still have a proper solutions afterward but not get an appropriate weight. This commit also adds some comments to the kernelsmoothing class.

### DIFF
--- a/DDECal/KernelSmoother.h
+++ b/DDECal/KernelSmoother.h
@@ -4,7 +4,19 @@
 #include <cmath>
 #include <stdexcept>
 #include <vector>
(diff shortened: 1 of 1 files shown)


aroffringa:
The second commit fixes  #128. @tammojan could you merge when agreed on the change?

PR #141, opened 2019-05-04
Unsignedint

tammojan:
Some changes (including a huge search-replace for `uint` -> `unsigned int`) to make DP3 compile on my Mac again. I disabled the build of the TestDynDPPP and PythonDPPP on Mac.


aroffringa:
Seems fine to me. At some point many of those `unsigned int`s should actually be `size_t` (whever it is an indexing var).

PR #142, opened 2019-05-06
Remove dependency on Boost signals 1

aroffringa:
New Boost version no longer provide Boost signals version 1. It has been replaced
by version 2, which is part of the Boost core library and no longer requires an
explicit find_package parameter.


aroffringa:
Oh this is already fixed in the idgpredict branched, so if that's merged this is also fixed (and the branches conflict).

aroffringa:
I'll merge this now and merge master again in idgpredict, because not being able to compile master makes things harder.

PR #143, opened 2019-05-06
Add parameter `antennaconstraint`

### LINKED ISSUE #140: Specify station names in coreconstraint
It would be useful to have the possibility to specify which station to constraint in the core constraint with a list of station names. Currently one can specify only "meters from the core". This would slightly increase the complexity of the parset for users but it would make the code more flexible allowing strategies where once a station converged to good solutions it is then blocked in further iterations.

aroffringa:
This allows direct specification of lists of antennas to force to have the same solution. Fixes #140 (if it works -- untested so far).

### DIFF
--- a/Common/StreamUtil.h
+++ b/Common/StreamUtil.h
@@ -28,9 +28,10 @@
 
 #include <map>
 #include <ostream>
+#include <set>
+#include <string>
 #include <utility>
 #include <vector>
-#include <string>
 
 namespace DP3
 {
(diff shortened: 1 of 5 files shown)


aroffringa:
I've tested this, and it works, so ready to be merged by @tammojan .

It turned out that the original coreconstraint didn't work with full Jones solutions, so I've also made the coreconstraint more generic so that it constrains with any nr of pols.

tammojan:
It might be nice to use the antenna selection syntax here. My impression is that that would be quite straightforward, if it isn't then never mind.

aroffringa:
Your probably right, but I don't know anything about that (I never understood that syntax...). If you think it's useful you're happy to change the code. Could also be done in a separate follow-up pull request.

PR #145, opened 2019-05-14
Enable tests

tammojan:
This PR enables two tests, and fixes an issue with nthreads initialization.

One question before merging: I probably did the CMake dependencies wrong, since now '`make -j`' gives problems if stuff is built in the wrong order. Should we fix this first?


aroffringa:
> This PR enables two tests, and fixes an issue with nthreads initialization.
> 
> One question before merging: I probably did the CMake dependencies wrong, since now '`make -j`' gives problems if stuff is built in the wrong order. Should we fix this first?

Yeah I think so. Shall I take a look at the cmake files and convert these tests to Boost testing?

tammojan:
If you can, please. I think the cmake and boost testing are two separate
issues, but if you can fix one or both, please do.

Op di 14 mei 2019 11:35 schreef André Offringa <notifications@github.com>:

aroffringa:
I've also added a more complex test; tDynStep. When running it, something goes wrong with the dynamic linking, which is the same reason that we linked aoflaggerstep etc. directly in. Hence, for now I've disabled it, but at least it compiles again.

I'm happy having this merged.

Issue #1, opened 2018-09-11
Replace assert statements with exceptions


aroffringa:
The code currently has many "assert(..)" statements, that in most cases seem to catch user errors (e.g. things like a missing parmkey value). assert statements should only be used for catching code bugs by the developer, since asserts are not enabled in a release build. When user errors happen, an exception should directly be thrown instead.

aroffringa:
Fixes in #159 .

PR #159, opened 2019-06-20
Replace asserts with throw std::runtime_error when appropriate

aroffringa:
Pull request to finish #1. I think the rule should be that any error caused by bad input
(parset error, command line error, data format error, etc) should throw an exception.
Any check on internal consistency or logic can be an assert(), so that they are disabled when
building in RELEASE mode. Some of the e.g. parset errors are now checked with 'assert',
which means that they won't be checked when build in RELEASE mode. Moreover, an assert
error is not very helpful for the user, an exception is at least a slightly friendlier
description of the problem.

### DIFF
--- a/Common/BaselineSelect.cc
+++ b/Common/BaselineSelect.cc
@@ -38,8 +38,6 @@
 #include <casacore/casa/Arrays/Matrix.h>
 #include <casacore/casa/Arrays/Vector.h>
 
-#include <cassert>
-
 using namespace casacore;
 
 namespace DP3 {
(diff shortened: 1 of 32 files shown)


tammojan:
Thanks, this looks good. I would like to rephrase some of the error messages; are you still working on this branch or could I fix it on this branch?

aroffringa:
I'm all in favour of rephrasing error messages :). I had done a few more changes ; I'll commit and push these to this branch and then stay off. I decided not to involve the demixing code, since they sprinkled with asserts that I don't fully understand, plus I'm not sure how alive that code is anyway...

aroffringa:
Oh crap, I committed something that doesn't compile; will you fix that too @tammojan ?

tammojan:
> Oh crap, I committed something that doesn't compile; will you fix that too @tammojan ?

I was about to find out :) Sure, fixed now.

tammojan:
@aroffringa Back to you; could you merge if it's ok with you?

aroffringa:
I had not yet gone through all the files ; shall I first do that and then merge everything at once?

tammojan:
Fine with me; I'll stay off the branch for now.

aroffringa:
@tammojan commit is finished now, ready to merge?

Fixes #1 .

Issue #144, opened 2019-05-06
Use antenna selection syntax where appropriate


tammojan:
Wherever an antenna is specified, it would be good to use the Antenna Selection Syntax (https://casacore.github.io/casacore-notes/263). For example in the Antenna constraint (#143).

revoltek:
As a quick workaround the antennaconstraint could ignore instead of return an error if an antenna doesn't exist. This would help the pipeline usage on datasets with different numbers of antennas.

tammojan:
Sorry for the late reply. I think you're referring to a different issue here, since the bug I was referring is merely a syntacical issue. Could you open a new issue to describe what you want to have fixed? Ideally with a reproducible error?

PR #161, opened 2019-06-25
Make antenna constraint work if antenna is missing

tammojan:
With this fix, constraints like `[[CS001, CS002, CS003], [RS301, RS401]]` will work even if `CS003` is not there. As suggested in #144.


tammojan:
Please check travis output, I haven't tested if this builds..

PR #166, opened 2019-07-09
Beam keywords when not inverting

aroffringa:
Jishnu was trying to use an applybeam step with `invert=false` and found this incorrectly reported:

```
std exception detected: In applying the beam: the metadata of
this observation indicate that the beam has already been applied
```

### DIFF
--- a/DPPP/ApplyBeam.cc
+++ b/DPPP/ApplyBeam.cc
@@ -119,10 +119,19 @@ namespace DP3 {
         itsDirection = MDirection(q0, q1, type);
       }
       
-      if(info().beamCorrectionMode() != NoBeamCorrection)
-        throw std::runtime_error("In applying the beam: the metadata of this observation indicate that the beam has already been applied");
-      info().setBeamCorrectionMode(itsMode);
-      info().setBeamCorrectionDir(itsDirection);
+      if(itsInvert)
+      {
+        if(info().beamCorrectionMode() != NoBeamCorrection)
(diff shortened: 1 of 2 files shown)


DPPP/ApplyBeam.cc
```diff
@@ -119,10 +119,19 @@ namespace DP3 {
...
+      if(itsInvert)
+      {
+        if(info().beamCorrectionMode() != NoBeamCorrection)
+          throw std::runtime_error("In applying the beam (with invert=true): the metadata of this observation indicate that the beam has already been applied");
+        info().setBeamCorrectionMode(itsMode);
+        info().setBeamCorrectionDir(itsDirection);
+      }
+      else {
+        if(info().beamCorrectionMode() == NoBeamCorrection)
+          throw std::runtime_error("In applying the beam (with invert=false): the metadata of this observation indicate that the beam has not yet been applied");
```
tammojan: I think this error should not be thrown. If you have a simulated data set without beam, you may want to apply (corrupt with) the beam later on.

aroffringa: Hmm not sure. When simulating a LOFAR observation without beam, it doesn't make sense to corrupt with the beam, as this would only corrupt the visibilities with a central LOFAR beam, which is not representative of a LOFAR observation, since the differential part is not applied. So you could say that the beam correction is only sensible in two cases: a) when processing a real LOFAR observation; and b) when processing a simulation that underwent full LOFAR beam application. I think in those two cases all should go well.

aroffringa:
BTW Jishnu tested the fix and it solves the `invert=false` exception.

Issue #168, opened 2019-07-10
Segfault when finishing DDE solve with subtract


aroffringa:
MS has 5393 timesteps and I'm solving with `solint=30` -- maybe that causes an error?

Output + stack trace:
```
0%....10....20....30....40....50....60....70....80....90....100%
Finishing processing ...

Program received signal SIGSEGV, Segmentation fault.
DP3::DPPP::DDECal::subtractCorrectedModel (this=this@entry=0x57b67a0, fullJones=true) at /home/users/offringa/Software/DP3/DDECal/DDECal.cc:1320
1320                  value +=

#0  0x0000000000619367 in DP3::DPPP::DDECal::subtractCorrectedModel(bool) (this=this@entry=0x57b67a0, fullJones=true)
    at /home/users/offringa/Software/DP3/DDECal/DDECal.cc:1320
#1  0x000000000061f0de in DP3::DPPP::DDECal::doSolve() (this=this@entry=0x57b67a0) at /home/users/offringa/Software/DP3/DDECal/DDECal.cc:781
#2  0x000000000061f834 in DP3::DPPP::DDECal::finish() (this=0x57b67a0) at /home/users/offringa/Software/DP3/DDECal/DDECal.cc:1262
#3  0x000000000052ca80 in DP3::DPPP::GainCal::finish() (this=0xd3a060) at /home/users/offringa/Software/DP3/DPPP/GainCal.cc:1329
#4  0x0000000000469c0b in DP3::DPPP::DPRun::execute(std::string const&, int, char**) (parsetName="calibrate-dde-new-splitted.parset", argc=argc@entry=2, argv=argv@entry=0x7fffffffc7e8) at /home/users/offringa/Software/DP3/DPPP/DPRun.cc:203
#5  0x000000000045b532 in main(int, char**) (argc=2, argv=0x7fffffffc7e8) at /home/users/offringa/Software/DP3/DPPP/DPPP.cc:86
```

Input parset:
```
msin=L254104_SAP000_SB044_uv.MS.5ch4s.dppp
msin.ntimes=10

msout=.
msout.datacolumn=CORRECTED_DATA

steps=[gaincal, ddecal]

gaincal.caltype = fulljones
gaincal.solint = 1
gaincal.nchan = 1
gaincal.applysolution = true
gaincal.sourcedb = models/field-new.sourcedb
gaincal.usebeammodel = true
gaincal.onebeamperpatch = false
gaincal.parmdb = di-solutions-new.h5

[...]

PR #169, opened 2019-07-10
Fix dde subtract one-off bug

aroffringa:
This bug (#168) caused DDECal to crash when `subtract=true` due to a one-off error in the use of `itsStepInSolInt`.

### DIFF
--- a/DDECal/DDECal.cc
+++ b/DDECal/DDECal.cc
@@ -830,7 +830,7 @@ void DDECal::doSolve ()
 
   itsTimer.stop();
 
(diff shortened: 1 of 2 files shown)


aroffringa:
Tested, seems to solve the issue.

As mentioned in the commit log, (unlike I mentioned to you irl) there isn't a bug in GainCal, just in DDECal subtract. This is because the counter that is somewhat inconsistently used, is not used inside the `calibrate()` function. It's not necessary, as GainCal relies on the data having been set to zero, and 'unset' data therefore doesn't affect the solutions.

Issue #167, opened 2019-07-09
segfault when pre-applying multiple h5parms


revoltek:
I get a segfault when I try to preapply 2 h5parms in a dde solve. Might it be because the time-resolution of the two h5parms I want to apply are different?

This is my command:

DPPP /home/fdg/scripts/LiLF/parsets/LOFAR_dd-incr/DPPP-solTEC.parset 
msin=mss-dd/TC08.MS 
sol.applycal.steps=[myapplycal1,myapplycal2] 
sol.applycal.myapplycal1.correction=tec000 
sol.applycal.myapplycal2.correction=tec000 
sol.applycal.myapplycal1.parmdb=ddcal/solutions/cal-core-c0.h5 
sol.applycal.myapplycal2.parmdb=ddcal/solutions/cal-remote-c0.h5 
sol.solint=10 sol.nchan=4 sol.h5parm=mss-dd/TC08.MS/cal-remote2-c0.h5 
sol.sourcedb=ddcal/skymodels/skymodel00_cluster.skydb

aroffringa:
Can you make a gdb stack backtrace?

tammojan:
Hi Francesco, could you point me to the data you're using?

tammojan:
Never mind, found them.

tammojan:
I have reproduced the segfault and am working on this.

tammojan:
For reference; the issue does not occur when using one of the two applycal's. The stacktrace points to HDF5:

```
#0  0x00007ffff77c0416 in H5SL_remove () at /lib64/libhdf5.so.8
#1  0x00007ffff769477e in H5FO_delete () at /lib64/libhdf5.so.8
#2  0x00007ffff763ff3d in H5D_close () at /lib64/libhdf5.so.8
#3  0x00007ffff77078bd in H5I_dec_ref () at /lib64/libhdf5.so.8
#4  0x00007ffff77079b5 in H5I_dec_app_ref () at /lib64/libhdf5.so.8
#5  0x00007ffff7707b83 in H5I_dec_app_ref_always_close () at /lib64/libhdf5.so.8
#6  0x00007ffff761b58e in H5Dclose () at /lib64/libhdf5.so.8
#7  0x00007ffff7bc0838 in H5::DataSet::close() () at /lib64/libhdf5_cpp.so.8
#8  0x00007ffff7bc2316 in H5::DataSet::~DataSet() () at /lib64/libhdf5_cpp.so.8
[...]

PR #173, opened 2019-07-17
Simplify locking for HDF5

tammojan:
There was a case where the HDF5 mutex was not set when calling ApplyCal
from nested threads. Rather than trying to pass the mutex around, I've
introduced a static mutex, which also cleans up a bit. This mutex is
used only for reading HDF5, so is not in any critical part of execution.

Solves #167

### DIFF
--- a/DPPP/Predict.cc
+++ b/DPPP/Predict.cc
@@ -434,10 +434,7 @@ namespace DP3 {
 
       // Call ApplyCal step
       if (itsDoApplyCal) {
-        if(itsMeasuresMutex == nullptr)
-          itsApplyCalStep.process(itsTempBuffer);
-        else
-          itsApplyCalStep.process(itsTempBuffer, itsMeasuresMutex);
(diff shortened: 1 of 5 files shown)


DPPP/Predict.cc
```diff
@@ -434,10 +434,7 @@ namespace DP3 {
 
       // Call ApplyCal step
       if (itsDoApplyCal) {
-        if(itsMeasuresMutex == nullptr)
-          itsApplyCalStep.process(itsTempBuffer);
-        else
-          itsApplyCalStep.process(itsTempBuffer, itsMeasuresMutex);
```
tammojan: I don't know why itsMeasuresMutex was used here; it seems that an existing threading mutex was reused for HDF5.

aroffringa: I might have done that, not sure though. Happy to blame you as well. In that case it's despicable!

PR #181, opened 2019-08-27
Fix180

tammojan:
I agree this introduces some code duplication, it's better though than the previous case where the baseline specification worked in the Filter step but not in the MSReader..


aroffringa:
Looks good to me. I've compiled and `make test` -ed the branch and that works as well.

PR #183, opened 2019-08-29
Add option msin.starttimeslot

tammojan:
This makes it possible to start processing at the n-th time slot.

### DIFF
--- a/DPPP/MSReader.cc
+++ b/DPPP/MSReader.cc
@@ -172,6 +172,10 @@ namespace DP3 {
       // They can also be used to select part of the MS.
 
       if (!startTimeStr.empty()) {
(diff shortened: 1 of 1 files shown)


DPPP/MSReader.cc
```diff
@@ -183,7 +185,7 @@ namespace DP3 {
         // Round specified first time to a multiple of itsTimeInterval
         itsFirstTime = startTimeMS + std::ceil((startTimeParset - startTimeMS) / itsTimeInterval) * itsTimeInterval;
```
aroffringa: In this case you should probably report an error when startTimeSlot != 0 (specifying a time + a startslot is not allowed)

PR #191, opened 2019-09-26
Refant

### LINKED ISSUE #190: No valid solutions if station 0 is flagged in tecandphase mode
DPPP does not produce valued solutions if the first station in the ms is fully flagged. This happens in ddecal tecandphase mode (in this case CS001HBA is 100% flagged). So somehow it is not allowed to have a fully flagged station 0 (almost like there is some referencing going with respect to station 0...).  After applying these h5 solutions (NaNs if I am correct) the entire ms gets flagged.

(this is DPPP is a relatively recent DPPP version in the singularity image make by @tikk3r, so ~few weeks old)

tammojan:
Fixes #190 

The algorithm to chose the reference antenna implemented here to choose the first antenna with at least 20% unflagged channels.

### DIFF
--- a/DDECal/TECConstraint.cc
+++ b/DDECal/TECConstraint.cc
@@ -69,6 +69,9 @@ void TECConstraintBase::applyReferenceAntenna(std::vector<std::vector<dcomplex> 
       // Choose this refAntenna;
       break;
   }
+  // All antennas are flagged, use first one (will lead to NaNs for this solint)
+  if (refAntenna == _nChannelBlocks)
+    refAntenna = 0;
 
   for(size_t ch=0; ch!=_nChannelBlocks; ++ch)
   {
--- a/DDECal/TECConstraint.cc
+++ b/DDECal/TECConstraint.cc
@@ -53,8 +53,21 @@ void ApproximateTECConstraint::initializeChild()
 
 void TECConstraintBase::applyReferenceAntenna(std::vector<std::vector<dcomplex> >& solutions) const
 {
-  // TODO chose this more cleverly?
+  // Choose reference antenna that has at least 20% channels unflagged
   size_t refAntenna = 0;
+  for(; refAntenna!=_nAntennas; ++refAntenna)
+  {
+    size_t nUnFlaggedChannels=0;
+    // Only check flagged state for first direction
+    for (size_t ch=0; ch!=_nChannelBlocks; ++ch)
+    {
+      if(isfinite(solutions[0][refAntenna*_nDirections]))
+        nUnFlaggedChannels++;
+    }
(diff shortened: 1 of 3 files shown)


DDECal/TECConstraint.cc
```diff
@@ -53,8 +54,21 @@ void ApproximateTECConstraint::initializeChild()
...
-  // TODO chose this more cleverly?
+  // Choose reference antenna that has at least 20% channels unflagged
   size_t refAntenna = 0;
+  for(; refAntenna!=_nAntennas; ++refAntenna)
+  {
+    size_t nUnFlaggedChannels=0;
+    // Only check flagged state for first direction
+    for (size_t ch=0; ch!=_nChannelBlocks; ++ch)
+    {
+      if(isfinite(solutions[0][refAntenna*_nDirections]))
```
aroffringa: Something is wrong here; `ch` is not used.

tammojan:
Showing the value of `doPhaseReference` in `DDECal::show()` is unfortunately not easy.

DDECal/TECConstraint.cc
```diff
@@ -53,8 +54,21 @@ void ApproximateTECConstraint::initializeChild()
...
+        nUnFlaggedChannels++;
+    }
+    if (nUnFlaggedChannels*1.0/_nChannelBlocks > 0.2)
+      // Choose this refAntenna;
+      break;
+  }
```
aroffringa: Add a check:
```
if(refAntenna == _nAntennas)
  throw std::runtime_error("Could not found a reference antenna: all antennas appear to be flagged");
```

aroffringa:
> Showing the value of `doPhaseReference` in `DDECal::show()` is unfortunately not easy.

If you want that, you can just store a copy of the bool in the DDECal class. Not sure we should however copy the entire parset's contents to stdout, which is what we are slowly moving toward. If people want to know the settings used, they can just cat the parset. Runtime parameters (like nr of ch) are more interesting.

DDECal/TECConstraint.cc
tammojan: Will change to `ch`

DDECal/TECConstraint.cc
tammojan: I don't think this is a good idea. When one timeslot is fully flagged, we don't want DPPP to crash on a runtime error.

aroffringa: Ah that's right, I forgot this is per solution interval. But an `if` is required nevertheless, because coming out with `refant=nAnt` will cause a segfault. So then it should just do `refant=0` in the case that no antenna is found.

tammojan: Good point, I did that.

PR #192, opened 2019-09-26
Fix compatibility with older HDF5

tammojan:
This fixes the build on CEP3. Inspired by debian bug 898567.


tammojan:
Thanks for testing.

PR #195, opened 2019-10-24
Make DP3 work with casacore tab64bit branch

aroffringa:
This is a draft branch to test compilation with casacore's tab64bit branch.

It contains a few TODO comments that need to change before merging. Also, it should not be merged before casacore's tab64bit branch is merged, because it won't compile with older versions of casacore.

### DIFF
--- a/DPPP/Filter.cc
+++ b/DPPP/Filter.cc
@@ -266,7 +266,7 @@ namespace DP3 {
       // Note that stations of baselines that have been filtered out before,
       // will also be removed.
       Vector<rownr_t> removedAnt = selTab.rowNumbers();
-      Vector<rownr_t> antMap = createIdMap (antTab.nrow(), removedAnt);
+      Vector<int> antMap = createIdMap (antTab.nrow(), removedAnt);
       antTab.removeRow (removedAnt);
       // Remove and renumber the stations in other subtables.
       Table ms(msName);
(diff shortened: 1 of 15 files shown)


DPPP/Filter.cc
```diff
@@ -265,12 +265,12 @@ namespace DP3 {
       // Remove these rows from the ANTENNA table.
       // Note that stations of baselines that have been filtered out before,
       // will also be removed.
-      Vector<uInt> removedAnt = selTab.rowNumbers();
-      Vector<Int> antMap = createIdMap (antTab.nrow(), removedAnt);
+      Vector<rownr_t> removedAnt = selTab.rowNumbers();
+      Vector<rownr_t> antMap = createIdMap (antTab.nrow(), removedAnt);
```
tammojan, line 269: This should not be a `rownr_t`, createIdMap uses a -1 for non-mapped antennas.

DPPP/Filter.cc
```diff
@@ -265,12 +265,12 @@ namespace DP3 {
       // Remove these rows from the ANTENNA table.
       // Note that stations of baselines that have been filtered out before,
       // will also be removed.
-      Vector<uInt> removedAnt = selTab.rowNumbers();
-      Vector<Int> antMap = createIdMap (antTab.nrow(), removedAnt);
+      Vector<rownr_t> removedAnt = selTab.rowNumbers();
```
tammojan, line 268: While this is a `rownr_t`, it maps to the antenna table, I think we can safely assume the number of antennas to fit in an int.

tammojan:
The tests will likely fail because the tab64bit branch of casacore has an old version. The code now checks for casacore version 3.3.0 and typedefs an internal `rownr_t` (in namespaces DPPP, BBS::ParmDB and BBS::SourceDB) to either casacore's `rownr_t` or `unsigned int`.

I've tested that DPPP, linked to 64bit casacore, can read a 32bit MS. The result can be read back with DPPP linking to the current casacore master (so without 64bit rownrs). I have not tested with an actual 64bit MS.

When this PR is merged, I think casacore/casacore#945 can be merged as well.

tammojan:
@aroffringa I've set travis to use the released version of casacore again, so that travis will not fail when the tab64 bit branch get merged.
I think this PR is ready to be merged when Travis passes.

aroffringa:
Great! Nice to have both versions of casacore work again. I'm happy to have it merged as well when Travis passes.

PR #198, opened 2019-11-13
Improve H5Parm error message

tammojan:
When doing a fulljones applycal with a H5Parm with only XX and YY
polarizations in, previously DP3 would crash with the HDF5 Error
"H5Dread(): selection+offset not within extent".
This patch makes it a bit more clear what goes wrong.


DPPP/SolTab.cc
```diff
@@ -256,6 +256,13 @@ namespace DP3 {
       nfreqH5 = getFreqIndex(freqs[freqs.size()-1])-startFreq+1;
       freqAxisH5 = vector<double>(fullFreqAxisH5.begin()+startFreq, fullFreqAxisH5.begin()+startFreq+nfreqH5);
     }
+    if (hasAxis("pol")) {
+      unsigned int npolH5 = getAxis("pol").size;
+      if (pol>npolH5-1) {
```
aroffringa: I'd do `pol >= npolH5`, but ok not really an issue.

PR #206, opened 2019-12-25
Functionality to support IDG multi-term prediction

### LINKED ISSUE #202: Support spectral structure in idg predict
IDG predict might be a solution to predicting large number of sources, but the current implementation in DP3 doesn't support prediction of sources with varying spectra (spectral index + higher terms, or 'joined channel'-like).

aroffringa:
This will allow DDECal using IDG and multiple terms.

Fixes #202 .

### DIFF
--- a/DDECal/DDECal.cc
+++ b/DDECal/DDECal.cc
@@ -306,10 +306,11 @@ void DDECal::initializeConstraints(const ParameterSet& parset, const string& pre
 
 void DDECal::initializeIDG(const ParameterSet& parset, const string& prefix)
 {
-  std::string
-    imageFilename = parset.getString(prefix + "idg.image"),
-    regionFilename = parset.getString(prefix + "idg.regions");
-  itsFacetPredictor.reset(new FacetPredict(imageFilename, regionFilename));
+  std::string regionFilename =
+    parset.getString(prefix + "idg.regions");
+  std::vector<std::string> imageFilenames =
+    parset.getStringVector(prefix + "idg.images");
+  itsFacetPredictor.reset(new FacetPredict(imageFilenames, regionFilename));
   itsFacetPredictor->PredictCallback = std::bind(&DDECal::idgCallback, this, std::placeholders::_1, std::placeholders::_2, std::placeholders::_3, std::placeholders::_4);
   itsDirections.resize(itsFacetPredictor->NDirections());
   for(size_t i=0; i!=itsDirections.size(); ++i)
--- a/IDGPredict/FacetImage.h
+++ b/IDGPredict/FacetImage.h
@@ -26,6 +26,7 @@ public:
 	{
 		_width = width;
 		_height = height;
+    _data.resize(spectralTerms);
     for(size_t i=0; i!=spectralTerms; ++i)
       _data[i].assign(width * height, value);
 	}
--- a/IDGPredict/FacetImage.h
+++ b/IDGPredict/FacetImage.h
@@ -8,50 +8,50 @@
 class FacetImage
 {
 public:
-	FacetImage() :
-		_data(),
-		_offsetX(0),
-		_offsetY(0)
-	{ }
-	
-	size_t Width() const { return _width; }
-	
-	size_t Height() const { return _height; }
-	
-	int OffsetX() const { return _offsetX; }
-	
-	int OffsetY() const { return _offsetY; }
-	
-	void Set(size_t width, size_t height, size_t spectralTerms, double value)
-	{
-		_width = width;
-		_height = height;
+  FacetImage() :
+    _data(),
+    _offsetX(0),
+    _offsetY(0)
+  { }
+  
+  size_t Width() const { return _width; }
+  
+  size_t Height() const { return _height; }
+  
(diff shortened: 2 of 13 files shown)


DDECal/DDECal.cc
```diff
@@ -306,10 +306,11 @@ void DDECal::initializeConstraints(const ParameterSet& parset, const string& pre
...
 void DDECal::initializeIDG(const ParameterSet& parset, const string& prefix)
 {
-  std::string
-    imageFilename = parset.getString(prefix + "idg.image"),
-    regionFilename = parset.getString(prefix + "idg.regions");
-  itsFacetPredictor.reset(new FacetPredict(imageFilename, regionFilename));
+  std::string regionFilename =
+    parset.getString(prefix + "idg.regions");
+  std::vector<std::string> imageFilenames =
+    parset.getStringVector(prefix + "idg.images");
```
tammojan: New usage, not backward compatible (but that's probably no problem for the few people that now use IDG in DPPP). Please document at https://www.astron.nl/lofarwiki/doku.php?id=public:user_software:documentation:ndppp

IDGPredict/FacetImage.h
```diff
@@ -22,14 +22,16 @@ class FacetImage
 	
 	int OffsetY() const { return _offsetY; }
 	
-	void Set(size_t width, size_t height, double value)
+	void Set(size_t width, size_t height, size_t spectralTerms, double value)
 	{
 		_width = width;
 		_height = height;
-		_data.assign(width * height, value);
+    _data.resize(spectralTerms);
```
tammojan: Please use consistent tabs or spaces (holds for entire PR).

IDGPredict/FacetImage.h
```diff
@@ -107,7 +116,7 @@ class FacetImage
 		return clippedFacet;
 	}
 	
-	ao::uvector<double> _data;
+	std::vector<ao::uvector<double>> _data;
```
tammojan: Maybe comment here what the vectors do (e.g. For each spectral term, uvector with ...)

IDGPredict/FacetPredict.h
```diff
@@ -81,34 +88,38 @@ class FacetPredict
     int buffersize = 256;
     idg::api::options_type options;
     IdgConfiguration::Read(proxyType, buffersize, options);
-    ao::uvector<double> data;
+    std::vector<ao::uvector<double>> data(_readers.size());
     _metaData.clear();
+    FitsReader& reader = _readers.front();
```
tammojan: If I understand correctly, you can now provide a fits file with spectral index (and possibly higher order terms). Maybe add an assert here that all sizes are equal? I think otherwise stuff will crash.

IDGPredict/FacetPredict.h
```diff
@@ -157,51 +172,83 @@ class FacetPredict
...
-        float
-          rotSin = sin(angle),
-          rotCos = cos(angle);
-        for(size_t p=0; p!=4; ++p) {
-          std::complex<float> s = values[ch*4 + p];
-          values[ch*4 + p] = std::complex<float>(
-            s.real() * rotCos  -  s.imag() * rotSin,
-            s.real() * rotSin  +  s.imag() * rotCos);
+        double frequency = _bands[dataDescId][ch];
+        double freqFactor = frequency / _refFrequency - 1.0;
```
tammojan: Is this the definition of spectral index used? If it is not very standard, it would be good to add this to either documentation or comments.

DDECal/DDECal.cc
aroffringa: Yes that was still missing, thanks, now added to the docs.

IDGPredict/FacetImage.h
aroffringa: Indeed that was a bit messy; most IDGPredict files still used tabs because I started those in a different project. I've ran them all through `expand` now so that they use spaces consistently.

IDGPredict/FacetPredict.h
aroffringa: Are you referring to whether the images have the same size? That's a good idea, I've now added checks here:
https://github.com/lofar-astron/DP3/blob/497395ff94bb9176bfca0063f69b048d3fc3ddc6/IDGPredict/FacetPredict.h#L44
The number of spectral terms is determined from the number of images in the parset.

IDGPredict/FacetPredict.h
aroffringa: I've added:

      // The "polynomial spectrum" definition is used, equal to the one e.g. used by WSClean in component outputs
      // (see https://sourceforge.net/p/wsclean/wiki/ComponentList/ ) and in text files when 'logarithmic SI' is false. 
      // The definition is:
      //   S(nu) = term0 + term1 (nu/refnu - 1) + term2 (nu/refnu - 1)^2 + ...

IDGPredict/FacetImage.h
aroffringa: Good idea

aroffringa:
@tammojan Thanks for the review! Happy to merge?

Issue #212, opened 2020-02-06
Crash using TECConstraint in DDECal


tammojan:
As reported by @revoltek. Using the latest master, the following parset causes a crash. Reproduced on a clean ubuntu 16.04 docker image with minimal requirements installed.

```
msin = forTJ1sec.MS
msin.datacolumn = DATA
msout = .

numthreads=1
steps=[sol]

sol.type = ddecal
sol.h5parm = test.h5
sol.maxiter = 300
sol.tolerance = 5e-4
sol.solint = 1
sol.nchan = 16
sol.stepsize = 0.2

sol.mode = tec # phaseonly, complexgain, tec, tecandphase
sol.propagatesolutions = True
sol.approximatetec = True
sol.maxapproxiter = 250
sol.approxtolerance = 1e-3
sol.usemodelcolumn = True
```

Below is the output of valgrind.

```
root@b2e81ea7da64:/data/fra# valgrind DPPP
==39361== Memcheck, a memory error detector
==39361== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
==39361== Using Valgrind-3.11.0 and LibVEX; rerun with -h for copyright info
==39361== Command: DPPP
==39361==
MSReader
  input MS:       /data/fra/forTJ1sec.MS
  band            0
  startchan:      0  (0)
  nchan:          976  (0)
  ncorrelations:  4
  nbaselines:     666
  first time:     2015/10/25/23:00:04
  last time:      2015/10/25/23:00:04
  ntimes:         1
  time interval:  4.00556
  DATA column:    DATA
  WEIGHT column:  WEIGHT_SPECTRUM
  autoweight:     false
DDECal sol.
  H5Parm:              test.h5
  solint:              1
  nchan:               16
  directions:          [[]]
  use model column:    true
  tolerance:           0.0005
  max iter:            300
  flag unconverged:    false
     diverged only:    false
  propagate solutions: true
       converged only: false
  detect stalling:     true
  step size:           0.2
  mode (constraints):  tec
  approximate fitter:  true
  only predict:        false
  subtract model:      false
MSUpdater msout.
  MS:             /data/fra/forTJ1sec.MS
  datacolumn:     DATA
  weightcolumn    WEIGHT_SPECTRUM
  Compressed:     no

  flush:          0

Processing 1 time slots ...

0%==39361== Invalid read of size 8
==39361==    at 0x699E6B: operator/=<double> (complex:1334)
[...]

PR #213, opened 2020-02-07
Fix incorrect refantenna check

aroffringa:
Possible fix for #212 , crash related to TECConstraint.
Bug was introduced in #191 . Bug should only lead to a crash when all antennas are >20% flagged, which is inconsistent with what Francesco reports in https://github.com/lofar-astron/DP3/issues/212#issuecomment-583316960 . Nevertheless, this commit fixes a bug, and I don't see any other issues in the refant calculation.

This bug was btw already more or less mentioned by me in my review of #191, apparently the fix didn't make it properly and it got merged unnoticed :( .


tammojan:
I confirm that this fixes the crash in #212. Thanks André!

revoltek:
I also confirm this fixed the problem, many thanks to both of you for the super quick fix!

PR #222, opened 2020-03-19
Make it possible to correct for beam a second time

### LINKED ISSUE #221: Correcting array factor away from pointing center
As in the telecon today, it would be nice to be able to correct for the array factor in a direction away from the delay or reference direction.

I see that #68 was similar to this, where the `direction` option was added. Prefactor already corrects for the beam at the pointing center (where array factor = 1 if I'm not mistaken). Attempting to take out just the array factor in another direction (on a phaseshifted dataset) results in

```
std exception detected: In applying the beam (with invert=true): the metadata of this observation indicate that the beam has already been applied
```

I split a couple timeslots and channels from a phaseshifted MS that already has the beam taken out at the original phase center by prefactor and uploaded it here: ftp://ftp.strw.leidenuniv.nl/pub/sweijen/test_beam.ms.tar

aroffringa:
This makes it e.g. possible to first apply the dipole beam to some direction and then the full beam when phased to another direction. This was requested by Frits. 

When `invert=true` (i.e. normal situation when applying beam to data), and the beam has already been applied to a certain direction, this is first undone and then the requested beam to the requested direction is applied. Some examples:
- Applying an `invert=true` beam to direction A, then again to direction A, is now accepted, and ends up with the beam in direction A (second run doesn't change anything).
- Applying an `invert=true` element beam to direction A, then applying an `invert=true` array beam to direction B, would end up with the *array beam* applied to direction B (i.e. element beam to direction A is undone).

Fully negating the beam having applied it is also possible (and was already possible). It requires an `invert=false` beam correction with the same direction and mode setting as was applied. The `invert=xxx` option is pretty confusing, and I'll file a ticket that we should make the beam setting such that we specify what we want as output, not what operation we want to do.

Fixes #221.

### DIFF
--- a/DPPP/ApplyBeam.cc
+++ b/DPPP/ApplyBeam.cc
@@ -65,6 +65,7 @@ namespace DP3 {
           itsUpdateWeights(parset.getBool(prefix + "updateweights", false)),
           itsDirectionStr(parset.getStringVector(prefix+"direction", std::vector<std::string>())),
           itsUseChannelFreq(parset.getBool(prefix + "usechannelfreq", true)),
+          itsModeAtStart(NoBeamCorrection),
           itsDebugLevel(parset.getInt(prefix + "debuglevel", 0))
     {
       // only read 'invert' parset key if it is a separate step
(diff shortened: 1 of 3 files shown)


aroffringa:
I filed #223 as a step towards making the specification easier.

PR #227, opened 2020-04-14
Make DP3 compile against arrays2 branch of casacore

aroffringa:
This only consists of explicitly including the dependencies for each file and adding 'std::' where necessary. These changes therefore compile therefore also fine against
the master of casacore, hence they are good to be merged right away.

### DIFF
--- a/DPPP/DemixWorker.cc
+++ b/DPPP/DemixWorker.cc
@@ -47,8 +47,8 @@
 #include <casacore/casa/Quanta/MVEpoch.h>
 #include <casacore/casa/Arrays/Vector.h>
 #include <casacore/casa/Arrays/Matrix.h>
+#include <casacore/casa/Arrays/Array.h>
 #include <casacore/casa/Arrays/ArrayMath.h>
-#include <casacore/casa/Arrays/ArrayIO.h>
 #include <casacore/casa/Arrays/MatrixMath.h>
 #include <casacore/casa/Arrays/MatrixIter.h>
 #include <casacore/casa/Containers/Record.h>
@@ -604,7 +604,8 @@ namespace DP3 {
       if (nrtmp == 0) {
         return 0;
       }
-      return GenSort<float>::kthLargest (tmp, nrtmp, (nrtmp-1)/2);
+      std::nth_element(tmp, tmp+(nrtmp-1)/2, tmp+nrtmp);
+      return tmp[(nrtmp-1)/2];
     }
 
     void DemixWorker::setupDemix (unsigned int chunkNr)
--- a/DPPP/GainCalAlgorithm.cc
+++ b/DPPP/GainCalAlgorithm.cc
@@ -147,7 +147,7 @@ namespace DP3 {
         double ginit=0.;
         bool ginitcomputed=false;
         for (unsigned int ant=0; ant<_nUn; ++ant) {
-          if (!isFinite(_g(ant,0).real()) ) {
+          if (!std::isfinite(_g(ant,0).real()) ) {
             if (!ginitcomputed && !_stationFlagged[ant%_nSt]) {
               // Avoid calling getAverageUnflaggedSolution for stations that are always flagged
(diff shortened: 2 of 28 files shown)


aroffringa:
There are changes to use `std::nth_element` instead of Casacore's GenSort versions.

DPPP/DemixWorker.cc
```diff
@@ -604,7 +604,8 @@ namespace DP3 {
       if (nrtmp == 0) {
         return 0;
       }
-      return GenSort<float>::kthLargest (tmp, nrtmp, (nrtmp-1)/2);
+      std::nth_element(tmp, tmp+(nrtmp-1)/2, tmp+nrtmp);
+      return tmp[(nrtmp-1)/2];
```
tammojan, line 608: Does this mean that `GenSort<T>::kthLargest` is missing from the new arrays implementation? This would be an API breakage to document, I'm fine with using `std::` here.

DPPP/GainCalAlgorithm.cc
```diff
@@ -147,7 +147,7 @@ namespace DP3 {
         double ginit=0.;
         bool ginitcomputed=false;
         for (unsigned int ant=0; ant<_nUn; ++ant) {
-          if (!isFinite(_g(ant,0).real()) ) {
+          if (!std::isfinite(_g(ant,0).real()) ) {
```
tammojan, line 150: In implementation files, I think using `using namespace std::` is fine as well.

ParmDB/Parm.cc
```diff
@@ -37,7 +37,7 @@ namespace BBS {
       itsParmId  (parmid)
   {}
 
-  Parm::Parm (ParmCache& cache, const string& name)
+  Parm::Parm (ParmCache& cache, const std::string& name)
```
tammojan, line 40: This file would benefit from using `using namespace std::` in terms of readability.

ParmDB/ParmSet.h
```diff
@@ -56,11 +56,11 @@ namespace BBS {
     ParmSet();
 
     // Add a parm for the given ParmDB. If not existing in the parmDB,
-    // it will be added to the ParmDB when its values are written.
+    // it will be added to the ParmDB when its values are writtn.
```
tammojan, line 59: Curious why you think this is an improvement ;)

DPPP/DemixWorker.cc
aroffringa: Thanks -- GenSort is not missing from arrays2, but rather it's part of Utilities. However, the sorting routines were no longer pulled in via #include <..Arrays/Array.h>, so it gave an error. I could have fixed it with an appropriate #include, but I think it's better to use std when possible.

DPPP/GainCalAlgorithm.cc
aroffringa: I don't object to a `using namespace`, but there are some overloads inside casacore with the same name as things in std (e.g. sqrt, cos, etc). In those cases it is better to use std:: in front of them. Therefore, as a rule I'd rather use an explicit std::, but if it gets crazy using namespace is fine (unless of course it is used in templates and the chosen overload depends on the type).

ParmDB/ParmSet.h
aroffringa: Haha I did not do this knowingly :). I'll revert it. Good catch :).

DPPP/GainCalAlgorithm.cc
aroffringa: Btw 'abs' always needs std:: in front for floating point; if that's not done, the compiler may select int abs(int), and truncates the float.

aroffringa:
Thanks for the prompt review! Yes, I've tested this against casacore arrays2. There's one compilation error remaining to switch to arrays2, something with the switch from uInt to size_t for IPosition's indexing. For now I've just changed the 'easy to fix' things, which fortunately is almost everything.

PR #230, opened 2020-05-11
Fix compilation with g++ 10

aroffringa:
g++ 10 is slightly stricter on what headers to include. Test with e.g. "CXX=/usr/bin/g++-10 cmake ../".

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,9 +44,14 @@ endif(IDGAPI_LIBRARIES AND IDGAPI_INCLUDE_DIRS)
 
 find_package(CFITSIO REQUIRED)
 
+find_package(PythonLibs 3 REQUIRED)
+find_package(PythonInterp REQUIRED)
+message(STATUS "Using python version ${PYTHON_VERSION_STRING}")
+
 #Prevent accidentally finding old BoostConfig.cmake file from casapy
 set(Boost_NO_BOOST_CMAKE ON)
-find_package(Boost COMPONENTS date_time filesystem python3 program_options system unit_test_framework REQUIRED)
+set(BOOSTPY ${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR})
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -44,9 +44,14 @@ endif(IDGAPI_LIBRARIES AND IDGAPI_INCLUDE_DIRS)
...
 
+find_package(PythonLibs 3 REQUIRED)
+find_package(PythonInterp REQUIRED)
+message(STATUS "Using python version ${PYTHON_VERSION_STRING}")
+
 #Prevent accidentally finding old BoostConfig.cmake file from casapy
 set(Boost_NO_BOOST_CMAKE ON)
-find_package(Boost COMPONENTS date_time filesystem python3 program_options system unit_test_framework REQUIRED)
+set(BOOSTPY ${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR})
+find_package(Boost COMPONENTS date_time filesystem python${BOOSTPY} numpy${BOOSTPY} program_options system unit_test_framework REQUIRED)
```
tammojan, line 54: Unfortunately the naming of boost_python3 is different across distributions. I've seen python3, python37, python-py37 etc. Not sure how to fix this properly.

tammojan: Anyhow I think this is unrelated to this pull request, so maybe to be solved in a different branch?

aroffringa: It's necessary to fix Travis ; The latest Docker Ubuntu image got upgraded, and that one no longer supports the "boost_python3" syntax. It does support "boost_python37" syntax, and so does Debian (supports both), so I think this is the best way for now.

tammojan: Ah ok, I hadn't noticed the master is broken. I'll merge this branch when travis finishes.

aroffringa:
LofarBeam was fixed now too. I've restarted the build and with the new LofarBeam, things compile, so I'll merge.

PR #235, opened 2020-05-27
Remove phase referencing in RotationAndDiagonalConstraint

tammojan:
The phase referencing here is not necessary and will fail if the
first station is flagged.


aroffringa:
Don't you want to make it optional first, through a DDECal parset option (and maybe disable it by default) ?

rvweeren:
I think disable by default is preferred, like we did for TEC (having a DDECal parset option might be useful to verify the behavior is correct).

tammojan:
Optional is an option yes. It'll take me some time to propagate the option here...

tammojan:
Yes I'll have to make it optional if only for the tests, otherwise there is no deterministic outcome for the test (which now [fails](https://travis-ci.org/github/lofar-astron/DP3/jobs/691667224#L12001)):
```
/8 Test #8: rotationconstraint.test_rotation_and_diagonal ...***Failed    0.05 sec

Running 1 test case...

/src/DDECal/test/tRotationConstraint.cc(76): error: in "rotationconstraint/test_rotation_and_diagonal": check near(constraint_result[0].vals[0], 0.) has failed

*** 1 failure is detected in the test module "DP3"
```

tammojan:
Argh, I was confused about phase referencing. It's actually about rotation referencing here..

DDECal/test/tRotationConstraint.cc
```diff
@@ -48,11 +49,13 @@ BOOST_AUTO_TEST_CASE( test_rotation )
...
 
-BOOST_AUTO_TEST_CASE( test_rotation_and_diagonal )
+BOOST_DATA_TEST_CASE( test_rotation_and_diagonal,
+                      boost::unit_test::data::make({ true, false }),
+                      doPhaseReference )
 {
   RotationAndDiagonalConstraint constraint;
   constraint.InitializeDimensions(1, 1, 1);
-  constraint.SetDoPhaseReference(true);
+  constraint.SetDoPhaseReference(doPhaseReference);
```
tammojan, line 58: So sorry, this should have been `doRotationReference`...

DDECal/test/tRotationConstraint.cc
```diff
@@ -48,11 +49,13 @@ BOOST_AUTO_TEST_CASE( test_rotation )
...
 
   vector<vector<complex<double> > > onesolution(1);
```
tammojan, line 60: To test the constraint, you'll need at least two solutions (this dimension is the number of antennas).

DDECal/test/tRotationConstraint.cc
mnijhuis-tos: Should I also update the parset key in DDECal.cc?

tammojan: Yes, but only for this particular constraint; for the existing constraints (tec etc) it should stay phasereference.

tammojan: So you should add a parset key `rotationreference`.

DDECal/DDECal.cc
```diff
@@ -291,9 +281,13 @@ void DDECal::initializeConstraints(const ParameterSet& parset, const string& pre
 #endif
       break;
     case GainCal::ROTATIONANDDIAGONAL:
-      itsConstraints.emplace_back(new RotationAndDiagonalConstraint());
+    {
+      auto constraintPtr = boost::make_unique<RotationAndDiagonalConstraint>();
+      constraintPtr->SetDoRotationReference(parset.getBool(prefix + "rotationreference", true));
```
tammojan, line 286: as per discussion above, the default should be false

docs/schemas/DDECal.yml
```diff
@@ -141,3 +141,17 @@ inputs:
   applycal&#46;*:
     type: 
     doc: ApplyCal sub-step, same as in Predict step. One can pass an h5parm with as many directions as set in "directions" and each direction model is corrupted accordingly `.`
+  phasereference:
+    default: true
+    type: bool
+    doc: Use phase referencing. This setting is only effective when mode=tec or mode=tecandphase `.`
```
tammojan, line 147: Could you explain the need for the backtick here? I think it's prettier when `mode=tec` is rendered as code.

mnijhuis-tos: The tools we use for generating the html pages require the `.` at the end. I've prettyfied the mode=... texts.

PR #238, opened 2020-05-28
build: fixup python search and support BLAS_LIBS, LAPACK_LIBS

drhpc:
This is a set of changes we carry at Universität Hamburg to get
the LOFAR toolchain built. I hope it can be included upstream, or
at least something equivalent.

Important points are:

-  Use current (less deprecated) cmake ways to detect Python 3.
-  Less redundancy in Python checking in CMakeLists at deeper level
   (Always try to build PythonDPPP as Python and Boost is required
   anyway.)
-  Add support for BLAS_LIBS and LAPACK_LIBS, as this is
   customary in HPC installations to choose the (optimized) BLAS
   implementation in an environment that offers multiple options.

The last point is repeated in other patches we carry for the LOFAR
stack. Maybe the common source for the FindCasacore code should
be patched instead?

The motivation for us is to get the build going while being able to select
python and BLAS/LAPACK implementations using environment modules.

Edit: Remove Boost::pythonXY motivation, as this is already in current DPPP,
the change there is only the way of finding Python 3.

### DIFF
--- a/CMake/FindCasacore.cmake
+++ b/CMake/FindCasacore.cmake
@@ -17,6 +17,8 @@
 #
 # Variables used by this module:
 #  CASACORE_ROOT_DIR         - Casacore root directory. 
+#  BLAS_LIBS                 - override BLAS library
+#  LAPACK_LIBS               - override LAPACK library
 #
 # Variables defined by this module:
 #  CASACORE_FOUND            - System has Casacore, which means that the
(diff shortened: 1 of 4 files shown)


drhpc:
The commit history is messed up now … I guess I'd have to make a new fork or wrestle with git to make it nice. But could anyone at least comment on the intent and if it would be welcome to have BLAS_LIBS usage, for example?

revoltek:
I was discussing today with @tammojan how to best handle this - sorry it took a while

drhpc:
Well … at least the pull request looks good now again. It's bad that you can mess up existing history in git. But then, you can delete distory and remake it. It's patch management, but not really revision control.

tammojan:
Thanks for this PR, it looks good to me. I'll try to review merge this week. Sorry for the late reply.

aroffringa:
Replaced by #275

PR #241, opened 2020-06-13
Check version of AOFlagger

aroffringa:
This is in preparation of AOFlagger's 3.0 release, for which its API will change. By checking
inside CMake, the user is presented with an AOFlagger version error instead of
compilation errors once we switch.


tammojan:
I think a more proper way to detect AOFlagger is that AOFlagger exports a CMake configuration, similar to [how IDG does this](https://git.astron.nl/RD/idg/-/blob/master/idg-api/CMakeLists.txt#L147). That way if AOFlagger needs a new library (say, hypothetically, lua) this does not require an update of DP3.

However for older AOFlagger versions the config is not exported, so the find script will still be necessary. It would be good if AOFlagger 3 exports the config, and the FindModule searches for a config file first.

aroffringa:
Hi Tammo Jan. I fully agree. Once things within aoflagger are more solid, we can move the findaoflagger script to the aoflagger source repos. However, we need a transitional period for people that use an old DP3 and install a newer AOFlagger. At least now everything is already in a separate file, so maybe not the end goal, but a step in the right direction :). 

Is it possible to make AOFlagger 3 export the config, and that it would override the cmake file in DP3/CMake ? Do you want any changes to what I currently committed?

basvandertol:
Vanaf CMake 3.15 kan de user CMAKE_FIND_PACKAGE_PREFER_CONFIG=True zetten
(https://cmake.org/cmake/help/v3.15/release/3.15.html#id12)
In eerdere versie moet expliciet NO_MODULE of CONFIG worden meegegeven aan find_package om niet de module te gebruiken.

tammojan:
Maybe you can use the construct such as in CMake's internal FindBoost ([here](https://github.com/Kitware/CMake/blob/master/Modules/FindBoost.cmake#L434)): first try to find the aoflagger config, and if it exists do not search further. That would make a future and backward compatible solid FindAOFlagger module.

tammojan:
@basvandertol I'm not in favor of upgrading the CMake version requirement to 3.15 yet. That version of CMake was only released in July 2019 and so is not available in stable releases of linux distributions.

tammojan:
> Is it possible to make AOFlagger 3 export the config, and that it would override the cmake file in DP3/CMake ? Do you want any changes to what I currently committed?

No, you would need to check for an exported config in DP3's FindAOFlagger. So that's the change I suggest (following the template of FindBoost mentioned above).

aroffringa:
@tammojan I don't understand exactly what you mean (nor am I a cmake expert) -- are you able to make that change?

CMake/FindAOFlagger.cmake
```diff
@@ -0,0 +1,105 @@
...
+    list(APPEND AOFLAGGER_LIB ${CFITSIO_LIBRARY})
+    list(APPEND AOFLAGGER_INCLUDE_DIR ${CFITSIO_INCLUDE_DIR})
+  elseif()
+    message(STATUS "AOFlagger dependency CFITSIO not found.")
+    set(_dependencies_met FALSE)
+  endif()
+
+  find_package(PkgConfig)
+  pkg_check_modules(GTKMM gtkmm-3.0>=3.0.0)
+  pkg_check_modules(SIGCXX sigc++-2.0)
```
mnijhuis-tos, line 76: Why do you check sigc++ ? It's not used in the remainder of this file.

I see it was in the original CMakeLists.txt, but there it's also unclear why sigc++ is there.

aroffringa: Thanks, good point. AOFlagger searches for it too and links with it when it is found. It's also part of GTKMM I believe, maybe requiring it changed over time. I've removed it now.

tammojan:
Unfortunately I'm not a CMake expert either. But I think the goal of exporting a config file is that AOFlagger itself stores in the exported config the libraries it needs to be linked. AOFlagger does not export its config yet; but if it does in the future the FindAOFlagger will read this config instead of guessing from the environment how AOFlagger was built.
I modeled the changes after FindBoost. Finding the config file can be disabled by setting `AOFLAGGER_NO_AOFLAGGER_CMAKE`.

aroffringa:
Fine by me like this, thanks.

PR #242, opened 2020-06-15
Fix doxygen documentation, add some missing copyright headers

mnijhuis-tos:


tammojan:
I think all the `//@ingroup NDPPP` can go, do you agree?

tammojan:
When merging, I'm in favor of reducing the number of commits here (I can do then when merging).

mnijhuis-tos:
> I think all the `//@ingroup NDPPP` can go, do you agree?

When I remove all those statements, including the /// \addtogroup NDPPP, the 'Modules' tab in the doxygen output does not show the NDPPP module anymore: That group simply doesn't exist anymore. For free functions in the group, doxygen also ignores the doxygen comments.

Navigating all DPPP classes and free functions will be difficult that way: The DP3/DPPP namespace has no doxygen, so doxygen does not generate a list of all namespace members either...

When I add "/// Foo" above one of the DPPP namespace definitions, doxygen does generate a list of all DPPP namespace members, and the free functions in the namespace get their descriptions back.

I would prefer using the existing namespace structure over defining a new NDPPP group. The only question that remains is where to put the DPPP namespace documentation.

aroffringa:
@mnijhuis-tos I think it is common to do that in a separate file, somethink like dppp-namespace.dox.

mnijhuis-tos:
An alternative to explicitly documenting namespaces is setting "EXTRACT_ALL = YES". The result looks good. The only drawback is that it disables the warnings for undocumented functions.

tammojan:
This all looks great to me. Could @mnijhuis-tos rebase the commits a bit to reduce the number of commits to ~5 (or one, also fine with me).

PR #245, opened 2020-06-17
Use binaries in Travis

aroffringa:
Since travis/Github supports Ubuntu 20.04 now, casacore and aoflagger are available from debian. This doesn't test DP3 against the latest casacore/DP3, but it *should* in any case be tested against stable releases as well, and this is a lot faster than building casacore+aoflagger by hand.


aroffringa:
Improves Travis build speed from  1 hr 10 min 29 sec to 13 min 40 sec  :). Ready to be merged, @tammojan or @mnijhuis-tos ?

tammojan:
Wow thanks! Think of all the trees you saved by this.

Good idea to keep the old dockerfiles. If you agree I'd like to move all docker files out of the .travis subdirectory into a docker subdirectory.

aroffringa:
@tammojan  Sure

docker/ubuntu_20_04_nolofar
```diff
@@ -0,0 +1,22 @@
...
+    casacore-dev aoflagger-dev
+
+ADD . /src
+WORKDIR /src
+
+RUN mkdir /build && cd /build && cmake ../src
+RUN cd /build && make -j2
+RUN cd /build && make install -j2
+RUN DPPP
+RUN cd /build && CTEST_OUTPUT_ON_FAILURE=1 make test -j2
```
lkrombeenc, line 22: It could be better to chain all these commands into a single line. This can reduce the size of your container because it limits the amount of layers that are built.

PR #248, opened 2020-06-19
Add the design document to Doxygen

lkrombeenc:


tammojan:
We should expand this document to link the 'Stages' section to the stages in the DPStep object. The design document should be way more detailed than the poster it's based on.

PR #252, opened 2020-06-19
Enable more tests

tammojan:


aroffringa:
This seems redundant, as these tests are included by newer PRs.

PR #254, opened 2020-06-24
Remove remaining warnings

mnijhuis-tos:
DPPP should compile without any warnings because of the definition of done.

### DIFF
--- a/PythonDPPP/CMakeLists.txt
+++ b/PythonDPPP/CMakeLists.txt
@@ -37,6 +37,9 @@ if(${Boost_PYTHON${BOOSTPY}_FOUND})
   
   find_package(Boost REQUIRED COMPONENTS python${BOOSTPY})
   include_directories(${Boost_INCLUDE_DIR})
+  # By default, Boost-Python has auto_ptr support, which triggers a compiler
+  # warning, since auto_ptr is deprecated -> Disable auto_ptr support.
+  add_definitions(-DBOOST_NO_AUTO_PTR)
 
 #  add_subdirectory(test)
 else()
--- a/PythonDPPP/CMakeLists.txt
+++ b/PythonDPPP/CMakeLists.txt
@@ -37,9 +37,6 @@ if(${Boost_PYTHON${BOOSTPY}_FOUND})
   
   find_package(Boost REQUIRED COMPONENTS python${BOOSTPY})
   include_directories(${Boost_INCLUDE_DIR})
-  # By default, Boost-Python has auto_ptr support, which triggers a compiler
-  # warning, since auto_ptr is deprecated -> Disable auto_ptr support.
-  add_definitions(-DBOOST_NO_AUTO_PTR)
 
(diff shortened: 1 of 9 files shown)


aroffringa:
I wouldn't remove parameter names just because they cause a warning when compiling with -Wextra. They can be important for the documentation of interfaces even if they are not used, which is indeed the case of most of your changes.

In general, I rather stick to -Wall ; there's a reason that -Wextra is extra ;).

aroffringa:
Also please don't compile with -Werror, older (or newer) compilers might warn for different things, which makes it all in all a bit less usable for the users.

PythonDPPP/CMakeLists.txt
```diff
@@ -37,6 +37,9 @@ if(${Boost_PYTHON${BOOSTPY}_FOUND})
   
   find_package(Boost REQUIRED COMPONENTS python${BOOSTPY})
   include_directories(${Boost_INCLUDE_DIR})
+  # By default, Boost-Python has auto_ptr support, which triggers a compiler
+  # warning, since auto_ptr is deprecated -> Disable auto_ptr support.
+  add_definitions(-DBOOST_NO_AUTO_PTR)
```
aroffringa, line 42: For what compiler / boost version is this necessary? auto_ptr was removed from c++14 I don't get any warnings with Boost, but Boost of course builds fine. If possible I'd rather keep Boost at its default settings, instead of adding extra options for old Boost versions.

docker/ubuntu_20_04_lofar
```diff
@@ -23,8 +23,7 @@ RUN mkdir /lofarbeam/build && cd /lofarbeam/build && cmake -DCMAKE_INSTALL_PREFI
 ADD . /src
 WORKDIR /src
 
-RUN mkdir /build && cd /build && cmake ../src
-RUN cd /build && make -j2
-RUN cd /build && make install -j2
+RUN mkdir /build && cd /build && cmake -DCMAKE_CXX_FLAGS_CUSTOM=-Werror ../src
```
aroffringa, line 26: For now, let's not enable error on Warnings in CI. If we find this leaves us with warnings that are a problem, we can enable it.

lkrombeenc: While we are working here, would it be an idea to change the 4 run commands to a single run command. Docker will create a layer for each RUN command resulting in a larger container, which feels unnecessary here.

aroffringa: @lkrombeenc fine to me, can you commit that change to this branch? (assuming @mnijhuis-tos is finished with the branch)

mnijhuis-tos: Note: This branch has no changes in the docker files anymore. I don't mind having extra improvements in this branch but you can also create a new branch.

aroffringa: Right, a new branch is probably better.

PR #256, opened 2020-06-26
Remove empty yaml keys; Fix ApplyBeam reference

lkrombeenc:
There are some mistakes in the documentation:
- https://www.astron.nl/citt/DP3/steps/GainCal.html#applybeamtomodelcolumn, there is a parameter in a parameter. This is caused by empty yaml values
- The links to ApplyBeam are incorrect since they are linked to an applybeam parameter instead of the step (e.g. https://www.astron.nl/citt/DP3/steps/GainCal.html#applybeamtomodelcolumn)

This PR fixes these issues.


docs/schemas/GainCal.yml
```diff
@@ -87,30 +87,23 @@ inputs:
     doc: >-
       Debugging. If debuglevel==1, then a file debug.h5 is created containing all iterands. This file will be very large; you can use it to check the convergence speed etc `.`
   sourcedb:
-    type: 
     doc: >-
       Same as in :ref:`Predict` step `.`
   sources:
-    type:
```
tammojan, line 94: This is a bug in the docs themselves, but the type in here should correspond to the type in the Predict step (in this case, `string?`. See https://github.com/lofar-astron/DP3/blob/master/docs/schemas/Predict.yml

docs/schemas/GainCal.yml
```diff
@@ -50,7 +50,7 @@ inputs:
     default: false
     type: bool
     doc: >-
-      Apply the beam model (at the phase center) to the visibilities in the model column. If this option is true, all options from :ref:`ApplyBeam` are valid as well (except .invert, since the model data will always be corrupted for the beam) `.`
+      Apply the beam model (at the phase center) to the visibilities in the model column. If this option is true, all options from `ApplyBeam <ApplyBeam.html>`__ are valid as well (except .invert, since the model data will always be corrupted for the beam) `.`
```
tammojan, line 53: I didn't know the syntax with trailing `__`, is that yaml? If it works, ok with me.

lkrombeenc: it is to define that the href is anonymous. Else it will give weird warnings it has similar hrefs

PR #259, opened 2020-07-01
Add BDABuffer and BDAIntervalBuffer, including tests

mnijhuis-tos:
(deleted old comment)


aroffringa:
I think the changes are good, but let's discuss how to proceed with branches etc. Since we might make a release branch, this might be merged to that as well instead.

DPPP/BDABuffer.cc
```diff
@@ -0,0 +1,216 @@
...
+    , weights_(weights)
+    , full_res_flags_(full_res_flags)
+    , uvw_{ uvw ? uvw[0] : std::numeric_limits<double>::quiet_NaN(),
+            uvw ? uvw[1] : std::numeric_limits<double>::quiet_NaN(),
+            uvw ? uvw[2] : std::numeric_limits<double>::quiet_NaN() }
+    {}
+
+    BDABuffer::BDABuffer(const std::size_t pool_size,
+                         const Fields& fields)
+    : data_()
```
aroffringa, line 55: This is not the google syntax for class member initialization (comma should be after variable), see https://google.github.io/styleguide/cppguide.html#Constructor_Initializer_Lists
but I hope the linting tool will correct this once we turn that on

DPPP/BDABuffer.cc
```diff
@@ -0,0 +1,216 @@
...
+                           const bool* const flags,
+                           const float* const weights,
+                           const bool* const full_res_flags,
+                           const double* const uvw)
+    {
+      if (!rows_.empty() && TimeIsLessEqual(time + interval, rows_.back().time_)) {
+        throw std::invalid_argument("Rows are not properly ordered");
+      }
+
+      const std::size_t kDataSize = n_channels * n_correlations;
```
aroffringa, line 149: To me it is a little bit confusing to name this like a formal constant. Google Style says "Variables declared constexpr or const, **and whose value is fixed for the duration of the program**, are named with a leading "k" followed by mixed case" -- since this is within a function, which is called multiple times maybe with different n_channels, I think this therefore does not satisfy that requirement, in line with how I would interpret a constant.

DPPP/BDABuffer.cc
```diff
@@ -0,0 +1,216 @@
...
+      float* row_weights = nullptr;
+      bool* row_full_res_flags = nullptr;
+
+      if (data_.capacity() > 0) {
+        row_data = data_.end();
+        if (data) {
+          data_.insert(data_.end(), data, data + kDataSize);
+        } else {
+          const float kNaN = std::numeric_limits<float>::quiet_NaN();
+          data_.insert(data_.end(), kDataSize, { kNaN, kNaN });
```
aroffringa, line 167: Als de data niet gelezen wordt, vul je dan alsnog de data maar met NaNs? Is dat niet een beetje verspilling -- kan data_ niet gewoon van grote nul blijven? (Idem met andere)

DPPP/BDABuffer.h
```diff
@@ -0,0 +1,243 @@
...
+      };
+
+    public:
+      /**
+       * Create a new BDABuffer.
+       * @param pool_size Size of the memory pool for this buffer.
+       *                  (number of items)
+       * @param fields The fields that should be enabled in this buffer.
+       */
+      explicit BDABuffer(std::size_t pool_size, const Fields& fields = Fields());
```
aroffringa, line 92: waarom explicit?

DPPP/BDABuffer.cc
mnijhuis-tos: When I read the style guide, I agree that kXxxx is only for fixed constants, not for 'variable' constants.

DPPP/BDABuffer.cc
mnijhuis-tos: Filling the data with NaN is not a waste in terms of space: That space was already allocated for the row.

When adding a now with nullptr for data, we could indeed simply set row.data_ to nullptr, too. Then you can have a situation where some rows in a buffer have data and others don't. Instead of a single GetNumberOfElements we would then also need different functions for each value type, since the amount of values may differ.

Always allocating data for a row also allows filling in the data for a row after adding it to the BDABuffer. I'm actually exploiting that feature in tBDAIntervalBuffer.cc. We can even write the data for multiple rows at once.

So yes, we could store nullptr in the row and force the user to always supply the row data upfront, but I think that restriction is not necessary. It will also complicate the interface.

DPPP/BDABuffer.h
mnijhuis-tos: If a constructor can be called with a single argument, using 'explicit' avoids implicit conversions. In this case, without 'explicit', a std::size_t can be implicitly converted to a BDABuffer when calling a function that expects a BDABuffer.

DPPP/BDABuffer.cc
aroffringa: Good explanation, sounds fine to me.

DPPP/BDABuffer.h
aroffringa: We might be talking about different constructors -- I'm talking about the BDABuffer(std::size_t pool_size, const Fields& fields = Fields()) constructor. For the single-parameter constructor I agree with explicit, for this one there's no use, is there? (I think it explicitly disallows calling with initializer list, like BDABuffer b = {size,Field}, but I don't see why it is necessary to disallow that)

mnijhuis-tos: Since `fields` has a default value, this constructor also accepts a single size_t parameter, so it is (also) a single-parameter constructor.

PR #260, opened 2020-07-03
Use aoflagger 3 interface with Lua strategies

aroffringa:


aroffringa:
This branch is ready to be merged in the development branch, so I've requested a review from @tammojan . It does break CI because the CI doesn't have aoflagger 3 yet.

PR #262, opened 2020-07-03
Migrate .cc unit tests to boost unit tests

mnijhuis-tos:


DPPP/UVWFlagger.cc
```diff
@@ -163,12 +163,13 @@ namespace DP3 {
...
         if (! itsCenter.empty()) {
           // A different phase center is given, so calculate UVW for it.
           NSTimer::StartStop ssuvwtimer(itsUVWTimer);
-          Vector<double> uvw = itsUVWCalc.getUVW (getInfo().getAnt1()[i],
-                                                  getInfo().getAnt2()[i],
-                                                  buf.getTime());
+          uvw = itsUVWCalc.getUVW (getInfo().getAnt1()[i],
+                                   getInfo().getAnt2()[i],
+                                   buf.getTime());
           uvwPtr = uvw.data();
```
mnijhuis-tos, line 173: Since a test crashed, I ran valgrind on the test and found usages of unitialized data: uvwPtr is set to uvw.data() but uvw goes out of scope.

DPPP/test/unit/tPSet.cc
```diff
@@ -0,0 +1,447 @@
+// tPSet.cc: Test program for class PreFlagger::PSet
```
mnijhuis-tos, line 1: Please use 'git diff -w master:DPPP/test/tPSet.cc boost-test:DPPP/test/unit/tPSet.cc' for a better diff of this file.
Main changes:
- Remove all 'using namespace xxx'. There was a name clash for cout, and the difference between casacore::Vector and std::vector is much clearer now.  The Google Style guide also forbids such using directives.
- Replace ASSERT by BOOST_CHECK
- Replace main() function by BOOST_AUTO_TEST_* statements.

DPPP/test/unit/tUVWFlagger.cc
```diff
@@ -0,0 +1,453 @@
+// tUVWFlagger.cc: Test program for class UVWFlagger
```
mnijhuis-tos, line 1: For a better diff, use 'git diff -w master:DPPP/test/tUVWFlagger.cc boost-test:DPPP/test/unit/tUVWFlagger.cc'.

DPPP/test/unit/tStationAdder.cc
```diff
@@ -544,36 +542,47 @@ void testPatterns()
...
+  test1( 10,  16, 32, 4, false);
 }
+
+BOOST_AUTO_TEST_CASE( test_add3 ) {
+  test2( 10,  16, 32, 4);
+}
+
+BOOST_AUTO_TEST_CASE( test_add4 ) {
+  // Unknown station.
+  //test3("{ns1:unknown, ns2:[cs01.s02, cs01.s01]}");
```
mnijhuis-tos, line 574: @lkrombeenc: We should at some point either enable or remove this test.

DPPP/UVWFlagger.cc
aroffringa: Well found

DPPP/test/integration/tNDPPP.cc
```diff
@@ -21,20 +21,23 @@
...
 #include <stdexcept>
 
+#include <boost/test/unit_test.hpp>
+
+#include "../../DPRun.h"
+
 using namespace DP3::DPPP;
 using namespace casacore;
 
+BOOST_AUTO_TEST_SUITE(ndppp)
```
aroffringa, line 39: NDPPP is the old name of the software; we should consistently use 'DPPP' now.

DPPP/test/unit/tApplyCalH5.cc
```diff
@@ -362,21 +362,48 @@ int main()
...
-    testampl(13, 3, true, false);
-    createH5Parm(vector<double>(), vector<double>());
-    testampl(9, 2, false, false);
-  } catch (std::exception& x) {
-    cout << "Unexpected exception: " << x.what() << endl;
-   return 1;
-  }
-  return 0;
+BOOST_AUTO_TEST_CASE( testampl2 ) {
+  vector<double> times;
```
aroffringa, line 370: Use initializer list?

DPPP/test/unit/tH5Parm.cc
```diff
@@ -1,40 +1,44 @@
...
 }
 
-int main(int, char**) {
+BOOST_AUTO_TEST_CASE( test_gridinterpolate ) {
   {
     size_t ntimes=7;
     {
       // Create a new H5Parm
-      cout<<"Create tH5Parm_tmp.h5"<<endl;
+      // cout<<"Create tH5Parm_tmp.h5"<<endl;
```
aroffringa, line 34: Most of these cout-statement can just be removed IMHO; maybe some can be changed into a comment to explain what's done

DPPP/test/unit/tPSet.cc
```diff
@@ -0,0 +1,447 @@
...
+class TestInput: public DP3::DPPP::DPInput
+{
+public:
+  TestInput(int nbl, int nchan, int ncorr)
+    : itsNChan(nchan), itsNCorr(ncorr)
+  {
+    info().init (itsNCorr, 0, itsNChan, 0, 0, 50, string(), string());
+    // Fill the baseline stations; use 4 stations.
+    // So they are called 00 01 02 03 10 11 12 13 20, etc.
+    casacore::Vector<casacore::Int> ant1(nbl);
```
aroffringa, line 50: We don't have to use casacore::Int-like types (also casacore::String,  uInt, etc.) -- they are an artefact (and, except for String, I expect are to be deprecated in casacore at some point). Where possible we also shouldn't use a casacore::Vector, because they are a) much slower than a std::vector; and b) have really confusing copy/reference semantics, but  we can save cleaning that up the latter for another time ;).

DPPP/test/unit/tPSet.cc
```diff
@@ -0,0 +1,447 @@
...
+  }
+private:
+  virtual bool process (const DP3::DPPP::DPBuffer&) { return false; }
+  virtual void finish() {}
+  virtual void show (std::ostream&) const {}
+
+  int itsNChan, itsNCorr;
+};
+
+} // end anonymous namespace
```
aroffringa, line 85: google style code says this should be
```
} // namespace
```

DPPP/test/unit/tPSet.cc
```diff
@@ -0,0 +1,447 @@
...
+        // Make sure the matrix is correct.
+        const casacore::Matrix<bool>& mat = pset.itsFlagBL;
+        BOOST_CHECK_EQUAL (mat.shape(), casacore::IPosition(2,4,4));
+        BOOST_CHECK ( mat(0,0) &&  mat(0,1) &&  mat(0,2) &&  mat(0,3));
+        BOOST_CHECK ( mat(1,0) &&  mat(1,1) &&  mat(1,2) &&  mat(1,3));
+        BOOST_CHECK ( mat(2,0) &&  mat(2,1) && !mat(2,2) && !mat(2,3));
+        BOOST_CHECK ( mat(3,0) &&  mat(3,1) && !mat(3,2) && !mat(3,3));
+      }
+      {
+        std::cout << "testBL 2" << std::endl;
```
aroffringa, line 135: Should the cout statements be removed? And maybe the explicit scoping should be replaced by separate unit tests (possible with fixture) ?

DPPP/test/unit/tPSet.cc
```diff
@@ -0,0 +1,447 @@
...
+        BOOST_CHECK_EQUAL (mat.shape(), casacore::IPosition(2,4,4));
+        BOOST_CHECK (!mat(0,0) &&  mat(0,1) &&  mat(0,2) && !mat(0,3));
+        BOOST_CHECK ( mat(1,0) && !mat(1,1) &&  mat(1,2) && !mat(1,3));
+        BOOST_CHECK ( mat(2,0) &&  mat(2,1) && !mat(2,2) &&  mat(2,3));
+        BOOST_CHECK (!mat(3,0) && !mat(3,1) &&  mat(3,2) && !mat(3,3));
+      }
+      // Some erronous ones.
+      std::cout << "testBL expected error 1" << std::endl;
+      bool err = false;
+      try {
```
aroffringa, line 167: try/catch statements can be removed

DPPP/test/unit/tStationAdder.cc
```diff
@@ -21,24 +21,28 @@
...
+
+#include "../../StationAdder.h"
+#include "../../DPInput.h"
+#include "../../DPBuffer.h"
+#include "../../DPInfo.h"
+#include "../../../Common/ParameterSet.h"
+#include "../../../Common/StringUtil.h"
+#include "../../../Common/StreamUtil.h"
+
+using namespace DP3;
```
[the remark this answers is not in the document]
aroffringa: Fully agree!

DPPP/test/integration/tNDPPP.cc
lkrombeenc: I agree that we should change this then, This PR is about the unit tests and splitting the integration tests from it to fix later. I propose we leave it as is and fix this when we fix all the integration tests.

DPPP/test/unit/tStationAdder.cc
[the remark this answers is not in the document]
lkrombeenc: I fixed most of them. Two `import casacore` remain because it cannot find the * operand for casacore::Complex. Any idea how to fix that?

DPPP/test/unit/tPreFlagger.cc
```diff
@@ -21,32 +21,36 @@
...
+#include "../../DPInfo.h"
+#include "../../../Common/ParameterSet.h"
+#include "../../../Common/StringUtil.h"
+
+using namespace DP3;
 using namespace DP3::DPPP;
 using namespace casacore;
 using namespace std;
 
+BOOST_AUTO_TEST_SUITE(preflagger)
```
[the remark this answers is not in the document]
lkrombeenc: If we do this we also need to rename all the methods and classes that are defined in the tests because they will violate unique naming constraints e.g. multiple execute() methods or TestInput class names.

DPPP/test/integration/tNDPPP.cc
aroffringa: Since this statement was added during these boost changes, I would suggest to immediately update it here. I agree we don't have to do it everywhere else in the code at this point. At many places the new name is already used, so we slowly move in that direction.

DPPP/test/unit/tPSet.cc
```diff
@@ -28,14 +28,10 @@
 #include <casacore/casa/Arrays/ArrayIO.h>
 #include <casacore/casa/Quanta/MVTime.h>
 
+#include <boost/make_unique.hpp>
 #include <boost/test/unit_test.hpp>
 
-using std::vector;
-using DP3::ParameterSet;
-using DP3::DPPP::DPInput;
-using DP3::DPPP::DPInfo;
```
lkrombeenc: Should we then not also remove the `#include "../../DPInfo.h"`?

DPPP/test/unit/tPSet.cc
```diff
@@ -277,58 +257,44 @@ namespace DP3 {
...
         BOOST_CHECK_EQUAL (pset.itsTimeSlot[2], 4u);
         BOOST_CHECK_EQUAL (pset.itsTimeSlot[3], 10u);
       }
-      // Some erronous ones.
-      bool err = false;
-      try {
-        ParameterSet parset;
+      {
+        DP3::ParameterSet parset;
         parset.add ("reltime", "[12:00:00]");
```
lkrombeenc: I think all of these blocks should be separate tests, especially the 3 cases of the exception cases since they are exactly the same except for one input

CMakeLists.txt
```diff
@@ -309,7 +310,6 @@ set(TEST_FILENAMES
   DPPP/test/unit/tScaleData.cc
   DPPP/test/unit/tStationAdder.cc
   DDECal/test/unit/tRotationConstraint.cc
-  DPPP_Interpolate/test/tInterpolateStep.cc
```
lkrombeenc: I think it is better to leave the test in de DPPP_Interpolate directory since we do the same for DDECal tests.

DPPP/test/unit/tPSet.cc
```diff
@@ -361,24 +395,31 @@ BOOST_AUTO_TEST_SUITE(pset)
...
 
-BOOST_AUTO_TEST_CASE( test_none ) {
-  TestPSet::testNone();
-}
+BOOST_AUTO_TEST_CASE( test_none ) { TestPSet().testNone(); }
 
-BOOST_AUTO_TEST_CASE( test_bl ) {
-  TestPSet::testBL();
-}
+BOOST_AUTO_TEST_CASE( test_bl1 ) { TestPSet().testBL1(); }
```
lkrombeenc, line 400: Cool you already fixed that :). Would it be possible to not have the tests in the class but just as unit tests. We should do that everywhere, but that is too big of a change but since we are reworking this one already we might as well do it

tammojan:
This PR removes some .tgz files from the repository and replaces them by their extracted files. IMHO this corrupts the repository, and creates several tens of MB to the repository. I'm checking if we can work around this by keeping the tgz and extracting it in CMake.

tammojan:
Merged as #264

PR #263, opened 2020-07-04
Boost test gcovr

jmmaljaars:
A first implementation of getting the coverage with `gcovr` in CI/CD. Since the created images are not pushed to a registry, the implementation in the `.gitlab-ci.yml` is a bit ugly. Do you have any suggestions @tammojan? (My bet would be to build a base image - with all the required dependencies installed - and build both a 'debug' or 'deployment'  version of DP3. The testing then can be done in the 'debug' image.)

Please note: there's no fancy capturing of the coverage output yet.


tammojan:
I'd prefer to have the coverage build in a separate CI stage, and out of the Dockerfile (since the Dockerfile also has a reason to exist apart from CI; it can be used by users to deploy DP3).

As for using a registry: I'm all for keeping intermediate containers. But since we are currently using only one runner, we probably don't even need to push it to a registry. I don't think using docker.io as a registry is the best solution, we may be able to use the built-in gitlab container registry.

jmmaljaars:
I certainly agree on having a separate coverage stage and with no traces of the code coverage in the Dockerfile. However, the way the CI/CD is currently being set-up doesn't easily lend itself to such a strategy,  since:
1. There is no registry right now, and re-using images from the build stage is therefore not possible in a later stage (as far as I know) 
2. So that means you have to rebuild an image with all the dependencies in place just for doing the code coverage. That feels a bit inefficient...

So maybe we should rethink the way we want to structure the CI/CD pipeline (as said above: I definitely would be in favor of building a base-image as a first step, and then reuse this base image to 1) compile a debug version of DP3 in order to do the testing, coverage, etc. and to 2) finally, build a deploy image).

lkrombeenc:
I am just trying some things at the moment. Will all be reverted if it is not working at the end of the day so we still have a clean branch :)

lkrombeenc:
After having difficulties getting it to work on an alpine base image I decided to revert the changes.

I agree with @jmmaljaars. We should use multiple docker files in a similar way docker multi-stage builds work. Not convinced if we should make a separate DP3 debug version to do the testing on instead of building it once as a "final" version, which we then use as another base image for the testing.

Lets discuss this tomorrow.

PR #265, opened 2020-07-08
Separate and optimize pipeline build steps; Add coverage to test step

lkrombeenc:


.gitlab-ci.yml
```diff
@@ -1,22 +1,41 @@
 stages:
+  - prepare
   - build
   - deploy-doc
 
 .env:
   before_script:
   - export DOCKER_IMAGE=`echo $CI_PROJECT_NAME | awk '{print tolower($0)}'`:$CI_COMMIT_SHORT_SHA
```
tammojan, line 8: This line is now outdated.

.gitlab-ci.yml
```diff
@@ -1,22 +1,41 @@
...
 
-build_lofar:
+build_base:
+  stage: prepare
+  extends: .env
+  script:
+  - docker build --tag dppp_base:20.04.0 -f ./docker/ubuntu_20_04_base .
```
tammojan, line 14: I think you'll need to add a commit hash to this tag. If two different branches of DPPP are building at the same time, they use the same docker registry.

tammojan: So maybe dppp_base_20.04.0:${CI_COMMIT_SHORT_SHA}

.gitlab-ci.yml
tammojan: So all extends env can go.

docker/ubuntu_20_04_base
```diff
@@ -1,5 +1,8 @@
 FROM ubuntu:20.04
 
+ADD . /src
```
tammojan, line 3: Why push this line to top? Now the base image can never be re-used (since the code always changes). I'd put this line on top of ubuntu_20_04_lofar.

.gitlab-ci.yml
jmmaljaars: Agreed, the more so since the base image won't be pushed to a registry. And thus it doesn't hurt if we have a unique (though ugly) tag

.gitlab-ci.yml
```diff
@@ -1,22 +1,41 @@
...
-  - docker build --tag ${DOCKER_IMAGE}_lofar -f ./docker/ubuntu_20_04_lofar .
+  - docker build --tag dppp_lofar:${CI_COMMIT_SHORT_SHA} -f ./docker/ubuntu_20_04_lofar .
 
-build_nolofar:
+build-test:
   stage: build
+  dependencies: 
+    - build_base
   extends: .env
+  image: dppp_base:20.04.0
```
jmmaljaars, line 29: Tag to `dppp_base` needs updating along with previous comment

.gitlab-ci.yml
```diff
@@ -1,22 +1,41 @@
...
-build_lofar:
+build_base:
+  stage: prepare
+  extends: .env
+  script:
+  - docker build --tag dppp_base:20.04.0 -f ./docker/ubuntu_20_04_base .
+
+build-lofar:
   stage: build
+  dependencies:
```
jmmaljaars, line 18: Like in `build-test`, you want to make use of the base image, right?

So there probably should be something: `image: dppp_base:{YOUR_UPDATED_TAG}`

.gitlab-ci.yml
lkrombeenc: I did not do that because that tag will also need to be automatically injected into the ubuntu_20_04_lofar file FROM line. I will look into that

docker/ubuntu_20_04_base
jmmaljaars: Agreed, the base image should not/need not contain DP3 related source code

docker/ubuntu_20_04_lofar
```diff
@@ -1,30 +1,8 @@
...
-ADD . /src
-WORKDIR /src
-
-RUN mkdir /build && cd /build && cmake ../src
-RUN cd /build && make -j2
-RUN cd /build && make install -j2
-RUN DPPP
-RUN cd /build && CTEST_OUTPUT_ON_FAILURE=1 make test -j2
+FROM dppp_base:20.04.0
+
```
jmmaljaars, line 2: In view of previous comment: here you want to `ADD` the `src/` directory (and set the `WORKDIR`)

.gitlab-ci.yml
tammojan: I think you need something like [this](https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact)

PR #266, opened 2020-07-08
Make DP3 compile with and depend on EveryBeam

aroffringa:
After merging this, the oskar branch should be deleted.


CMakeLists.txt
```diff
@@ -57,13 +54,15 @@ set(BOOSTPY ${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR})
...
+find_path(EVERYBEAM_INCLUDE_DIR NAMES EveryBeam/load.h)
+if(NOT EVERYBEAM_INCLUDE_DIR)
+  message(FATAL_ERROR "EveryBeam include file (EveryBeam/load.h) not found: please install https://git.astron.nl/RD/EveryBeam")
+endif()
+find_library(EVERYBEAM_LIB everybeam)
+if(EVERYBEAM_INCLUDE_DIR AND EVERYBEAM_LIB)
+  include_directories(${EVERYBEAM_INCLUDE_DIR})
+  message(STATUS "EveryBeam library found.")
+# TODO everybeam is no longer optional, so these files can become unconditional
   set(LOFAR_DEPENDENT_FILES
```
jmmaljaars, line 68: Is this list complet @aroffringa ? Found a couple more EveryBeam dependencies, i.e.:

```
#       modified:   DPPP/ApplyBeam.cc
#       modified:   DPPP/DemixWorker.cc
#       modified:   DPPP/MSReader.cc
#       modified:   DPPP/Predict.cc
```

aroffringa: Do you mean the LOFAR_DEPENDENT_FILES list? DPPP/MSReader.cc was not dependent on the presence of the beam, so I don't understand your question. These files can move to the regular target files.

jmmaljaars:
Compiling DP3 against EveryBeam seems to work fine on CI (and yes, it does work indeed). Compiling and testing locally (on DAS5), however, leads to a range of memory errors/segfaults. It seems that those segfaults are related to EveryBeam (and its predecessor LOFARBeam)

Any ideas/suggestions on moving forward @aroffringa @mnijhuis-tos ?

CMakeLists.txt
```diff
@@ -76,15 +75,8 @@ if(LOFAR_STATION_RESPONSE_DIR AND LOFAR_STATION_RESPONSE_LIB)
     tPredict
```
mnijhuis-tos, line 75: When LofarBeam/EveryBeam is not optional anymore, please also move the names of these tests near the spot where 
LOFAR_DEPENDENT_INTEGRATION_TESTS is used.

mnijhuis-tos:
> Compiling DP3 against EveryBeam seems to work fine on CI (and yes, it does work indeed). Compiling and testing locally (on DAS5), however, leads to a range of memory errors/segfaults. It seems that those segfaults are related to EveryBeam (and its predecessor LOFARBeam)
> 
> Any ideas/suggestions on moving forward @aroffringa @mnijhuis-tos ?

Valgrind shows a duplicate delete call in the boost unit test library. This error also occurs with the old version. When LOFARBeam is not used, the application seems to run fine, however, valgrind still shows the duplicate delete call in boost. Upgrading boost to 1.73 on das5 does not have any effect.

I think we'll need a separate task for analyzing this issue further. We could try creating an empty test case, which only has BOOST_CHECK(true). If it still occurs, try removing parts from DP3 until the error disappears.

CMakeLists.txt
```diff
@@ -56,13 +53,15 @@ set(BOOSTPY ${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR})
...
+find_path(EVERYBEAM_INCLUDE_DIR NAMES EveryBeam/load.h)
+if(NOT EVERYBEAM_INCLUDE_DIR)
+  message(FATAL_ERROR "EveryBeam include file (EveryBeam/load.h) not found: please install https://git.astron.nl/RD/EveryBeam")
+endif()
+find_library(EVERYBEAM_LIB everybeam)
+if(EVERYBEAM_INCLUDE_DIR AND EVERYBEAM_LIB)
+  include_directories(${EVERYBEAM_INCLUDE_DIR})
+  message(STATUS "EveryBeam library found.")
+# TODO everybeam is no longer optional, so these files can become unconditional
   set(LOFAR_DEPENDENT_FILES
```
aroffringa, line 65: Now that the beam files are no longer optional, the ```set(LOFAR_DEPENDENT_FILES ...)``` should be removed and these files should directly be added to the list of files belonging to the target

PR #269, opened 2020-07-09
Add BDA Buffer tests

lkrombeenc:


DPPP/test/unit/tBDABuffer.cc
```diff
@@ -0,0 +1,391 @@
+// tScaleData.cc: Test program for class ScaleData
```
mnijhuis-tos, line 1: Please update.

DPPP/test/unit/tBDABuffer.cc
```diff
@@ -0,0 +1,391 @@
...
+#include <boost/test/unit_test.hpp>
+#include <boost/test/data/test_case.hpp>
+
+#include "../../BDABuffer.h"
+
+using DP3::DPPP::BDABuffer;
+
+BOOST_AUTO_TEST_SUITE(bdabuffer)
+
+BOOST_AUTO_TEST_CASE( test_bda_initialization )
```
mnijhuis-tos, line 33: You don't have to use 'bda' in a testname for a test in the 'bdabuffer' testsuite. I would even some 'test' from the test names, since it's clear that it's a test.

DPPP/test/unit/tBDABuffer.cc
```diff
@@ -0,0 +1,391 @@
...
+    float weights[] {1};
+
+    buffer.AddRow(0., 1., 0, 0, n_channels, n_correlations, &data);
+    buffer.AddRow(0., 1., 0, 0, n_channels, n_correlations, nullptr, flags);
+    buffer.AddRow(0., 1., 0, 0, n_channels, n_correlations, nullptr, nullptr, weights);
+
+    BDABuffer buffer_copy {buffer};
+
+    BOOST_CHECK(buffer.GetData() != buffer_copy.GetData());
+    BOOST_CHECK_EQUAL(*buffer.GetData(), *buffer_copy.GetData());
```
mnijhuis-tos, line 59: You're only comparing the first data element here. Please augment this test:
- Test that the number of elements in the copy is correct.
- Test that all individual elements are equal in both buffers.
- Test that you cannot add any rows to the copy.

DPPP/test/unit/tBDABuffer.cc
```diff
@@ -0,0 +1,391 @@
...
+    const size_t n_channels {2};
+    const size_t n_correlations {3};
+    BDABuffer buffer {3 * n_channels * n_correlations + 1};
+    bool flags[] {true};
+    const std::complex<float> data {1};
+    float weights[] {1};
+
+    buffer.AddRow(0., 1., 0, 0, n_channels, n_correlations, &data);
+    buffer.AddRow(0., 1., 0, 0, n_channels, n_correlations, nullptr, flags);
+    buffer.AddRow(0., 1., 0, 0, n_channels, n_correlations, nullptr, nullptr, weights);
```
mnijhuis-tos, line 54: You are supplying pointers to single elements to AddRow. However, AddRow expects n_channels * n_correlations elements behind each pointer!
Please adjust the sizes of the input data. For example, use:
const bool kFlags[n_channels * n_correlations] = { true };

DPPP/test/unit/tBDABuffer.cc
```diff
@@ -0,0 +1,391 @@
...
+}
+
+BOOST_AUTO_TEST_CASE( test_bda_add_null_data )
+{
+    const size_t n_channels {2};
+    const size_t n_correlations {4};
+    BDABuffer buffer {n_channels * n_correlations, true, false, false, false};
+    BOOST_CHECK(buffer.AddRow(0., 1., 0, 0, n_channels, n_correlations));
+    BOOST_CHECK(std::isnan((*buffer.GetData()).real()));
+    BOOST_CHECK(std::isnan((*buffer.GetData()).imag()));
```
mnijhuis-tos, line 105: Please first check that GetData() does not return nullptr. Also in the rest of this test: Check pointer validity before using them.

DPPP/test/unit/tBDABuffer.cc
```diff
@@ -0,0 +1,391 @@
...
+    BOOST_CHECK_EQUAL(buffer.GetFlags(), nullptr);
+    BOOST_CHECK_EQUAL(buffer.GetWeights(), nullptr);
```
mnijhuis-tos, line 107: Please also check the fullResFlags, also in other tests.

DPPP/test/unit/tBDABuffer.cc
mnijhuis-tos, line 105: Please check all elements, instead of only the first one. Also applies to the tests below.

DPPP/test/unit/tBDABuffer.cc
```diff
@@ -0,0 +1,391 @@
...
+    BOOST_CHECK_EQUAL(buffer.GetNumberOfElements(), size_t {0});
+    BOOST_CHECK_EQUAL(buffer.GetData(), nullptr);
+    BOOST_CHECK_EQUAL(buffer.GetFlags(), nullptr);
+    BOOST_CHECK_EQUAL(buffer.GetWeights(), nullptr);
+    BOOST_CHECK_EQUAL(buffer.GetFullResFlags(), nullptr);
+}
+
+BOOST_AUTO_TEST_CASE( test_copy )
+{
+    const size_t n_channels {2};
```
mnijhuis-tos, line 45: Google style says that contants are written as kNChannels, for example.

DPPP/test/unit/tBDABuffer.cc
```diff
@@ -0,0 +1,391 @@
...
+    BDABuffer buffer {2 * n_channels * n_correlations};
+    const double uvw[] {1, 2, 3};
+    const double uvw2[] {3, 4, 5};
+    const double uvw3[] {5, 6, 7};
+
+    BOOST_CHECK(buffer.AddRow(0., 1., 0, 0, n_channels, n_correlations, nullptr, nullptr, nullptr, nullptr, uvw));
+    BOOST_CHECK(buffer.AddRow(0., 1., 0, 0, n_channels, n_correlations, nullptr, nullptr, nullptr, nullptr, uvw2));
+    BOOST_CHECK(!buffer.AddRow(0., 1., 0, 0, n_channels, n_correlations, nullptr, nullptr, nullptr, nullptr, uvw3));
+    BOOST_CHECK_EQUAL(buffer.GetNumberOfElements(), size_t {2 * n_channels * n_correlations});
+    const std::vector<BDABuffer::Row> rows = buffer.GetRows();
```
mnijhuis-tos, line 213: Why do you copy 'rows'? Using a reference is preferred here.

DPPP/test/unit/tBDABuffer.cc
```diff
@@ -0,0 +1,391 @@
...
+    BOOST_CHECK_EQUAL(rows[0].GetDataSize(), n_channels * n_correlations);
```
mnijhuis-tos, line 214: I'd say this check belong to another test: This test is only about uvw values.

DPPP/test/unit/tBDABuffer.cc
```diff
@@ -0,0 +1,391 @@
...
+    BOOST_CHECK_EQUAL(rows[0].uvw_[1], uvw[1]);
+    BOOST_CHECK_EQUAL(rows[1].uvw_[0], uvw2[0]);
+    BOOST_CHECK_EQUAL(rows[1].uvw_[1], uvw2[1]);
+}
+
+BOOST_AUTO_TEST_CASE( test_bda_clear )
+{
+    const size_t n_channels {2};
+    const size_t n_correlations {3};
+    BDABuffer buffer {3 * n_channels * n_correlations + 1};
```
mnijhuis-tos, line 225: Why do you add 1?

DPPP/test/unit/tBDABuffer.cc
```diff
@@ -0,0 +1,391 @@
...
+    const std::complex<float> data {1};
+    const float weights[] {1};
+
+    BOOST_CHECK(buffer.AddRow(0., 1., 0, 0, n_channels, n_correlations, &data));
+    BOOST_CHECK(buffer.AddRow(0., 1., 0, 0, n_channels, n_correlations, nullptr, flags));
+    BOOST_CHECK(buffer.AddRow(0., 1., 0, 0, n_channels, n_correlations, nullptr, nullptr, weights, fr_flags));
+    BOOST_CHECK(!buffer.AddRow(0., 1., 0, 0, n_channels, n_correlations, &data));
+    BOOST_CHECK_EQUAL(buffer.GetNumberOfElements(), size_t {3 * n_channels * n_correlations});
+    buffer.Clear();
+    BOOST_CHECK_EQUAL(buffer.GetNumberOfElements(), size_t {0});
```
mnijhuis-tos, line 237: You can also check that Clear() does not affect the buffer capacity, by calling AddRow until it fails.

DPPP/test/unit/tBDABuffer.cc
```diff
@@ -0,0 +1,391 @@
...
+    const bool fr_flags[] {false};
+    const std::complex<float> data {1};
+    const float weights[] {1};
+
+    buffer.AddRow(0., 1., 0, 0, n_channels, n_correlations, &data, flags, weights, fr_flags);
+
+    BOOST_CHECK_EQUAL(*buffer.GetData(size_t {0}), data);
+    BOOST_CHECK_EQUAL(*buffer.GetFlags(size_t {0}), flags[0]);
+    BOOST_CHECK_EQUAL(*buffer.GetFullResFlags(size_t {0}), fr_flags[0]);
+    BOOST_CHECK_EQUAL(*buffer.GetWeights(size_t {0}), weights[0]);
```
mnijhuis-tos, line 255: - GetXXX() may return nullptr, so check it.
- Check all elements instead of only the first one.
- Is the size_t cast really necessary here? Using 0 or 0u should work I think.

DPPP/test/unit/tBDABuffer.cc
```diff
@@ -0,0 +1,391 @@
...
+    buffer.AddRow(0., 1., 0, 0, n_channels, n_correlations, nullptr, nullptr, nullptr, flags);
+    
+    bool new_flags[] {true, false, true};
+    *buffer.GetFullResFlags() = new_flags[0];
+    *(buffer.GetFullResFlags() + 1) = new_flags[1];
+    *(buffer.GetFullResFlags() + 2) = new_flags[2];
+    BOOST_CHECK_EQUAL(*buffer.GetFullResFlags(), new_flags[0]);
+    BOOST_CHECK_EQUAL(*(buffer.GetFullResFlags() + 1), new_flags[1]);
+    BOOST_CHECK_EQUAL(*(buffer.GetFullResFlags() + 2), new_flags[2]);
+}
```
mnijhuis-tos, line 324: These tests mainly test that GetXXX()  returns the same pointer when called multiple times. I wouldn't go that far with testing, so please remove these tests.

DPPP/test/unit/tBDABuffer.cc
```diff
@@ -0,0 +1,391 @@
...
+
+BOOST_AUTO_TEST_CASE( test_time_is_near_equal ) {
+    BDABuffer buffer {0};
+    BOOST_CHECK_EQUAL(buffer.TimeIsEqual(9.999999991, 10.0000000), true);
+}
+
+BOOST_AUTO_TEST_CASE( test_time_is_not_near_equal ) {
+    BDABuffer buffer {0};
+    BOOST_CHECK_EQUAL(buffer.TimeIsEqual(9.99999999, 10.0000000), false);
+}
```
mnijhuis-tos, line 389: The Time* functions are static, so you don't need to create a BDABuffer in these tests.
Simply BDABuffer::TimeIsEqual(...) also works.

Also: Personally I prefer BOOST_CHECK(X) over BOOST_CHECK_EQUAL(X, true) and BOOST_CHECK(!X) over BOOST_CHECK_EQUAL(X, false).

DPPP/test/unit/tBDABuffer.cc
lkrombeenc: Feels like a bit too much for me to do this, since the test will fail anyways. Or is that come c++ related thing because then we should definitely do it :)

DPPP/test/unit/tBDABuffer.cc
lkrombeenc: For variables in methods it is optional so we can keep it this way

DPPP/test/unit/tBDABuffer.cc
lkrombeenc: Check that is will not add something if the space is not enough. So it got 1 pool left, but needs 6 so I expect it to do nothing. Will add comments

DPPP/test/unit/tBDABuffer.cc
lkrombeenc: Similar as above, I think this is overkill since the test should fail anyway, unless it is a c++ thing that must be done.

DPPP/test/unit/tBDABuffer.cc
mnijhuis-tos: If you don't check it, the entire test application will probably crash when using the null pointer. There's no NullPointerException in C++, which can be caught: The application just dies. It will thus never generate JUnit output with a list of failed and passed tests.

If you do check it, only that test will fail and the test executable can continue with other tests and it can generate JUnit output at the end.

DPPP/BDABuffer.cc
```diff
@@ -82,7 +83,8 @@ namespace DP3 {
     , weights_(other.weights_)
     , full_res_flags_(other.full_res_flags_)
     , rows_()
-    , remaining_capacity_(0)
+    , original_capacity_(other.original_capacity_)
+    , remaining_capacity_(other.original_capacity_)
```
mnijhuis-tos, line 87: The remaining capacity of the copy should remain 0.

DPPP/BDABuffer.cc
```diff
@@ -109,6 +111,7 @@ namespace DP3 {
         flags_it += kDataSize;
         weights_it += kDataSize;
         full_res_flags_it += kDataSize;
+        remaining_capacity_ -= kDataSize;
```
mnijhuis-tos, line 114: For a copy, the remaining_capacity is simply 0.

DPPP/BDABuffer.cc
```diff
@@ -82,7 +83,8 @@ namespace DP3 {
     , weights_(other.weights_)
     , full_res_flags_(other.full_res_flags_)
     , rows_()
-    , remaining_capacity_(0)
+    , original_capacity_(other.original_capacity_)
```
mnijhuis-tos, line 86: The capacity of the copy will typically be smaller than the original: Copying a std::vector (the memory pool) does not copy its capacity: It only copies the contents. I'll have to check the behavior for ao::uvector, but my intention was that the copy has no remaining capacity.
You can use the loop over the rows, below, for initializing original_capacity_.

DPPP/test/unit/tBDABuffer.cc
mnijhuis-tos: 'data' still only has a single element.

DPPP/BDABuffer.cc
lkrombeenc: I would expect that if you have a buffer that still has 12 pools left that a copy would also have 12 pools left. This is because I'd expect that calls on copies are similar to calls on the original one.

Calling AddRow on the original will succeed and on the copy will return false contradicts this.

However, this is my opinion on how copy/clone should work based on experience in Java. Could be that it is different in c++ in which case I would like to hear it as well and then we can fix it :)

DPPP/BDABuffer.h
```diff
@@ -36,6 +38,9 @@ namespace DP3 {
     class BDABuffer
     {
     public:
+      enum class Field { kData, kFlags, kWeights, kFullResFlags, kBitCount };
+      using Fields = EnumBitset<Field>;
```
aroffringa, line 42: The use of a EnumBitset to simulate a struct of boolean is too complicated. Why not just use struct Fields { bool hasData, hasFlags, hasWeights } ... etc ? (Or even inline these booleans in the main class). This adds lots of lines of code that people have to figure out to understand the semantics of a EnumBitset...

DPPP/BDABuffer.h
```diff
@@ -196,7 +193,7 @@ namespace DP3 {
       }
 
       static constexpr bool TimeIsEqual(double x, double y) {
-        return abs(x - y) < kTimeEpsilon;
+        return fabs(x - y) < kTimeEpsilon;
```
aroffringa, line 196: (std::abs is fine, see https://stackoverflow.com/questions/3118165/when-do-i-use-fabs-and-when-is-it-sufficient-to-use-stdabs , in either case add std::)

DPPP/BDAIntervalBuffer.cc
```diff
@@ -93,11 +90,7 @@ namespace DP3 {
       }
 
       // Create the result buffer and fill it.
-      auto result = boost::make_unique<BDABuffer>(pool_size,
-                                                  enable_data,
-                                                  enable_flags,
-                                                  enable_weights,
-                                                  enable_full_res_flags);
+      auto result = boost::make_unique<BDABuffer>(pool_size, fields);
```
aroffringa, line 93: As per google style code, only use `auto` with reason (e.g. templated code), not to prevent typing

DPPP/test/unit/tScaleDataBDA.cc
```diff
@@ -71,8 +75,8 @@ DPInfo GenerateDPInfo(int ntime, int nbl, int nchan, int ncorr) {
   info.init (ncorr, 0, nchan, ntime, 0., 5., string(), string());
   // Fill the baseline stations; use 2 stations.
   // So they are called 00 01 02 03 10 11 12 13 20, etc.
-  Vector<Int> ant1(nbl);
-  Vector<Int> ant2(nbl);
+  vector<int> ant1(nbl);
```
aroffringa, line 78: great idea to use vector instead of casacore::Vector :)

DPPP/BDAIntervalBuffer.cc
mnijhuis-tos: From the Google C++ style guide: "For example, you can assume that the return type of make_unique<Foo>() is obvious, but the return type of MyWidgetFactory() probably isn't."
The C++ Core Guidelines rule C.150 also say that for make_unique, 'auto' is better.

DPPP/BDABuffer.h
mnijhuis-tos: Since abs and fabs are not constexpr, I already replaced them by `((x > y) ? (x - y) : (y - x))`.

DPPP/BDAIntervalBuffer.cc
aroffringa: Ok, sounds like you're right!

aroffringa: Oh by the way, we don't use (and shouldn't use) all of the C++ Core Guidelines rules. I thought you were quoting the Google style guide. C150 is about make_unique, not about auto. Herb Sutter promotes using auto everywhere, which is conflicting with the (more pragmatically) google style that we follow, as are many of the other C++ core guidelines. make_unique is C++14 btw, and should not be used.

DPPP/BDABuffer.h
mnijhuis-tos: - The semantics of EnumBitset are nearly equal to those of std::bitset: Instead of figuring out the semantics of a the Fields struct, you can use your existing knowledge about std::bitset.
- A separate struct Fields with 4 booleans easily gets equally complicated as the EnumBitset: You'll want a constructor for default initialization, and a Set() function for setting all bits. Only the operator[] members are not needed.
- An EnumBitset can be reused, since it's a template. Using separate structs will cause duplicate code.

PR #275, opened 2020-07-21
fixup python search and support BLAS_LIBS, LAPACK_LIBS 

aroffringa:
Supersedes #238


aroffringa:
I'll merge this for now, but some of it will require further changes once we have also switched DP3 to use pybind11.

PR #278, opened 2020-07-22
Merge and simplify formatting scripts.

mnijhuis-tos:
This change significantly reduces the formatting script size while still maintaining full functionality.


jmmaljaars:
I can see the point why you want to simplify this - for the specific dp3 case at least. Nevertheless I wonder whether we actually would like to do this, since:
1. originally, the `run-clang-format.sh` script could be added to the pre-commit hook to do the reformatting when committing. In your implementation, you actually would do the format check if you add the script to the pre-commit hook. I wouldn't be in favor of that.
2. You restricted the applicability to `*.h` and `*.cc` files only, and deleted the option to exclude certain directories from being reformatted. That's probably ok for the DP3 repo, but the formatting scripts are intended to be re-used across multiple repositories, e.g. `wsclean` (with `*.h` and `*.cpp` files) and `everybeam` - with a couple of `externals` that you definitely don't want to reformat.

I can live with your edits regarding 2, but I wouldn't opt for your choices regarding 1, i.e. I am still in favor of splitting it up in a `run` and a `check` (of course, it's perfectly fine with me if you do the `check` entirely in the CI/CD `yml` file)

scripts/run-clang-format.sh
```diff
@@ -1,73 +1,53 @@
...
+# To hook this script to pre-commit include the line
+# "./scripts/run-clang-format.sh" to .git/hooks/pre-commit
+# and make sure pre-commit is an executable shell script.
 
-programname=$0
+#Script configuration for this repo. Adjust it when copying to a different repo.
 
-set -e
+#The directories that contain source files, which clang-format should format.
+SOURCE_DIRS=($(dirname "$0")/..)
```
jmmaljaars, line 16: I would opt for one `SOURCE_DIR` rather than providing the option to specify a list of `SOURCE_DIRS`. Since `find` will work recursively, there doesn't seem much of a use case for specifying a list of directories. 

What's more: if you have only one `SOURCE_DIR` it's convenient to `cd` into this directory before executing `find`. This allows to provide relative paths to the root of `SOURCE_DIR` in `EXCLUDE_DIRS`

scripts/run-clang-format.sh
```diff
@@ -1,73 +1,53 @@
...
-    echo "Script to format C++ header and source files. Expects a .clang-format file in current directory or parent"
-    echo ""
-    echo "Usage: $programname [-i filename/pattern] [-s skip directory]"
-    echo "  -i      include file(pattern), pass wild-card arguments as literals!"
-    echo "  -s      skip directory in diff. Must be specified as a path, e.g. ./external"
-    echo "  -h      display help"
-    exit 1
-}
+#Directories that must be excluded from formatting.
+EXCLUDE_DIRS=()
```
jmmaljaars, line 18: `EXCLUDE_DIRS` now need to be specified relative to the location of the script, that's a bit inconvenient. Can be easily fixed by opting for just one `SOURCE_DIR`, and `cd` into `SOURCE_DIR` before executing `find`

scripts/run-clang-format.sh
```diff
@@ -1,73 +1,53 @@
...
-INC_FILE=`sed "s/'//g"<<<$INC_FILE`
+# Convert EXCLUDE_DIRS into "-path dir1 -prune -o -path dir2 -prune -o ..."
+FIND_EXCLUDES=
+for e in ${EXCLUDE_DIRS[*]}; do
+  FIND_EXCLUDES+="-path $e -prune -o "
+done
 
-if [ ${#exclude_directories[@]} -eq 0 ]; then
-    EX_DIR=""
+find ${SOURCE_DIRS[*]} $FIND_EXCLUDES -type f \( $FIND_NAMES \) \
```
jmmaljaars, line 42: `cd` into source dir just above this line

PR #280, opened 2020-07-23
Update CFITSIO detection and usage (for DAS5)

mnijhuis-tos:


CMake/FindCFITSIO.cmake
```diff
@@ -32,6 +32,10 @@
 
 if(NOT CFITSIO_FOUND)
 
+  if(NOT CFITSIO_ROOT_DIR AND DEFINED ENV{CFITSIO_ROOT_DIR})
+    set(CFITSIO_ROOT_DIR "$ENV{CFITSIO_ROOT_DIR}")
+  endif()
```
mnijhuis-tos, line 37: DAS5 defines a nice CFITSIO_ROOT_DIR environment variable which was unused until now: We manually had to add -DCFITSIO_ROOT_DIR=... to the cmake invocation. With this change, that's no longer necessary.

CMakeLists.txt
```diff
@@ -42,6 +42,7 @@ else(IDGAPI_LIBRARIES AND IDGAPI_INCLUDE_DIRS)
 endif(IDGAPI_LIBRARIES AND IDGAPI_INCLUDE_DIRS)
 
 find_package(CFITSIO REQUIRED)
+include_directories(${CFITSIO_INCLUDE_DIRS})
```
mnijhuis-tos, line 45: For some reason the include directory for cfitsio was never added...

PR #281, opened 2020-07-23
Add BDAAverager step (time averaging only)

mnijhuis-tos:


tammojan:
Three requests:
* UVW averaging should not take weights into acount (as in normal DPPP Averager)
* Data and weights averaging should take weights into account
* Rename some variables to make it easier to read (as discussed).

The logic do determine averaging factors will become more complicated once this step matures.

PR #282, opened 2020-07-24
Aocommon submodule

jmmaljaars:
Add `aocommon` as a git submodule to `external/aocommon` and delegate a couple of DP3 utilities to `aocommon`.  This induces quite a number of changes at critical points, so it deserves a careful review.


DDECal/MultiDirSolver.cc
```diff
@@ -17,13 +17,16 @@
...
 
+#include <aocommon/matrix2x2.h>
+
 #ifdef AOPROJECT
 #include "ParallelFor.h"
 #else
-#include "../Common/ParallelFor.h"
+#include <aocommon/parallelfor.h>
+
+using namespace aocommon;
```
aroffringa, line 29: Avoid 'using namespace'; just alias the specific classes used, e.g.
```using aocommon::MC2x2;```
etc.

PR #283, opened 2020-07-27
Use dry run in git hook and on CI

mnijhuis-tos:


jmmaljaars:
Make sure you have the correct EveryBeam includes (https://github.com/lofar-astron/DP3/pull/279)

scripts/run-clang-format.sh
```diff
@@ -41,15 +48,23 @@ for e in ${EXCLUDE_DIRS[*]}; do
 done
 
 cd $SOURCE_DIR
-find . $FIND_EXCLUDES -type f \( $FIND_NAMES \) \
-  -exec clang-format -i -style=file \{\} +
+FILES=$(find . $FIND_EXCLUDES -type f \( $FIND_NAMES \) -print)
```
jmmaljaars, line 51: Why do you need `-print` option here?

scripts/run-clang-format.sh
```diff
@@ -25,8 +25,15 @@ SOURCE_EXT=(*.cc *.h)
 
 set -e
 
+# Detect run environment.
+if [ -n "$CI" ]; then
+  DRYRUN=" (dry run on CI)"
+elif [ -n "$GIT_AUTHOR_DATE" ]; then
+  DRYRUN=" (dry run in git hook)"
+fi
+
```
jmmaljaars, line 34: I assume that 
- the `$CI` variable is (automatically) non-empty when the script is invoked from a CI/CD pipeline ?
- `$GIT_AUTHOR_DATE` is non-empty when the script is invoked from a pre-commit hook? 
- and neither of these two variables exist if the script is invoked as stand-alone?

mnijhuis-tos: Exactly! I did not add any further comments, as the code already describes it. The $CI variable is documented at https://docs.gitlab.com/ee/ci/variables/predefined_variables.html . I found out $GIT_AUTHOR_DATE myself, by examining the environment when running a git commit hook.

scripts/run-clang-format.sh
mnijhuis-tos: When I do a "find . -path ./DPPP -prune -o -name *.cc" inside the DP3 directory, the output includes the ./DPPP directory, without any filename. When I add "-print", the output only includes the found *.cc files, and not that pruned directory name.

mnijhuis-tos: Even with "-type f", the output still has the pruned directory name if I don't add "-print".

PR #292, opened 2020-08-05
BdaAverager: Configure using parset

mnijhuis-tos:


DPPP/BDAAverager.cc
```diff
@@ -64,9 +64,13 @@ void BDAAverager::BaselineBuffer::Clear() {
...
 
-BDAAverager::BDAAverager(double bl_threshold_time, double bl_threshold_channel)
-    : bl_threshold_time_(bl_threshold_time),
-      bl_threshold_channel_(bl_threshold_channel),
+BDAAverager::BDAAverager(const DP3::ParameterSet& parset,
+                         const std::string& prefix)
+    : bl_threshold_time_(parset.getDouble(prefix + "timethresholdlength", 0.0)),
+      bl_threshold_channel_(
+          parset.getDouble(prefix + "freqthresholdlength", 0.0)),
+      max_interval_(parset.getDouble(prefix + "maxinterval", 3600.0)),
```
tammojan, line 72: Even though an hour is long, I think the cleaner solution would be to set the default at 0, which means 'do not enforce a maximum'.

docs/schemas/BDAAverager.yml
```diff
@@ -0,0 +1,42 @@
...
+      any averaging over frequencies. The number of channels for a baseline
+      becomes baseline length / freqthresholdlength * original number of
+      channels, rounded up.
+      The minchannels setting specifies the minimum number of channels.
+      The maximum numbef of channels is the original number of channels `.`
+    default: 0.0
+  maxinterval:
+    type: int
+    doc: >-
+      The maximum interval duration (in seconds), when averaging over time.
```
tammojan, line 34: What happens if this is not a multiple of the integration time?

mnijhuis-tos: Effectively the maximum interval duration is rounded down to the nearest multiple of the integration time. I'll add a comment.

PR #296, opened 2020-08-06
Prepare structures for BDA channel data.

mnijhuis-tos:


DPPP/DPInfo.cc
```diff
@@ -70,35 +74,114 @@ void DPInfo::init(unsigned int ncorr, unsigned int startChan,
...
+  }
+
+  // Check that all baselines have equal channel layouts.
+  for (std::size_t bl = 1; bl < itsChanFreqs.size(); ++bl) {
+    if (itsChanFreqs.front() != itsChanFreqs[bl] ||
+        itsChanWidths.front() != itsChanWidths[bl] ||
+        itsResolutions.front() != itsResolutions[bl] ||
+        itsEffectiveBW.front() != itsEffectiveBW[bl]) {
+      return false;
+    }
```
mnijhuis-tos, line 167: @tammojan : Should we also apply the 1 kHz tolerance (see below) to these comparisons?

tammojan:
When the issues above are resolved, I'm fine with merging this.

PR #303, opened 2020-08-31
Remove chunked reading

lkrombeenc:


DPPP/MSBDAReader.cc
```diff
@@ -174,52 +174,45 @@ bool MSBDAReader::process(const DPBuffer&) {
...
-  auto uvw = uvw_col.getColumnCells(cell_range);
-  auto time = time_col.getColumnCells(cell_range);
-  auto interval = interval_col.getColumnCells(cell_range);
-  auto exposure = exposure_col.getColumnCells(cell_range);
-  auto data_desc_id = data_desc_id_col.getColumnCells(cell_range);
-
-  unsigned i = 0;
-  while (nread_ < ms_.nrow() && i < info().nbaselines()) {
-    const double ms_time = time[i];
+  while (nread_ < ms_.nrow() && buffer->GetRemainingCapacity() > 0) {
```
mnijhuis-tos, line 184: - `buffer->GetRemaininingCapacity() > 0` makes little sense: If the remaining capacity is 1, AddRow will probably not succeed, as a row needs n_channels * n_correlations capacity.
- Removing 'i' as the loop variable is possible, but then you should create some loop end variable. For example:
```
auto read_end = std::min(ms_.nrow(), nread + info().nbaselines());
for (; nread_ < read_end; ++nread_) {
```

DPPP/MSBDAReader.cc
```diff
@@ -174,52 +174,45 @@ bool MSBDAReader::process(const DPBuffer&) {
...
+    const double ms_time = ScalarColumn<double>(ms_, "TIME")(nread_);
```
mnijhuis-tos, line 185: You can use `time_col(nread_)` here, instead of creating a new ScalarColumn.

DPPP/MSBDAReader.cc
```diff
@@ -174,52 +174,45 @@ bool MSBDAReader::process(const DPBuffer&) {
...
-      ++i;
       ++nread_;
       continue;
     }
 
     const auto ant12 = std::make_pair(ant1_col(nread_), ant2_col(nread_));
 
-    auto* data_ptr = read_vis_data_ ? data[i].data() : nullptr;
+    Complex* data_ptr = nullptr;
+    std::vector<Complex> data;
```
mnijhuis-tos, line 197: For clarity, please add `casacore::` to Complex, so it's clear that it's not a std::complex.

DPPP/MSBDAReader.cc
```diff
@@ -174,52 +174,45 @@ bool MSBDAReader::process(const DPBuffer&) {
...
+    if (read_vis_data_) {
+      data = data_col.get(nread_).tovector();
+      data_ptr = data.data();
+    }
+    Cube<float> weights = weights_col.get(nread_);
+    Cube<double> uvw = uvw_col.get(nread_);
+    double interval = interval_col(nread_);
+    double exposure = exposure_col(nread_);
+    int data_desc_id = data_desc_id_col(nread_);
```
mnijhuis-tos, line 206: All ScalarColumns (including time) can still benefit from the cell_range optimization, so please only de-optimize the ArrayColumns. You could then also re-introduce `i` again, and restore the old loop.

mnijhuis-tos (changes requested):

DPPP/MSBDAReader.cc
```diff
@@ -209,12 +203,19 @@ bool MSBDAReader::process(const DPBuffer&) {
...
 
     const auto ant12 = std::make_pair(ant1_col(nread_), ant2_col(nread_));
+    casacore::Complex* data_ptr = nullptr;
+    std::vector<casacore::Complex> data;
+    if (read_vis_data_) {
+      data = data_col.get(nread_).tovector();
+      data_ptr = data.data();
+    }
+    Cube<float> weights = weights_col.get(nread_);
+    Cube<double> uvw = uvw_col.get(nread_);
```
mnijhuis-tos, line 213: Optional: You could perform the conversion from the casacore::Cube to std::vector here, instead of below(in the AddRow() call).

mnijhuis-tos (approved):

PR #304, opened 2020-09-04
Pointer and facetpredict updates

mnijhuis-tos:


Common/buffered_lane.h
```diff
@@ -109,8 +110,9 @@ class lane_read_buffer {
   lane_read_buffer(const lane_read_buffer&) = delete;
   lane_read_buffer& operator=(const lane_read_buffer&) = delete;
 
-  Tp* _buffer;
-  size_t _buffer_size, _buffer_pos, _buffer_fill_count;
+  aocommon::UVector<Tp> _buffer;
```
tammojan, line 113: Good change I think, however is this tested? Also, since this is in aocommon, we should push it back to where it came from.

DDECal/DDECal.cc
```diff
@@ -306,11 +306,13 @@ void DDECal::initializeIDG(const ParameterSet& parset, const string& prefix) {
...
   std::vector<std::string> imageFilenames =
       parset.getStringVector(prefix + "idg.images");
-  itsFacetPredictor.reset(new FacetPredict(imageFilenames, regionFilename));
-  itsFacetPredictor->PredictCallback = std::bind(
-      &DDECal::idgCallback, this, std::placeholders::_1, std::placeholders::_2,
-      std::placeholders::_3, std::placeholders::_4);
-  itsDirections.resize(itsFacetPredictor->NDirections());
+  itsFacetPredictor = boost::make_unique<FacetPredict>(
+      imageFilenames, regionFilename,
+      [this](std::size_t row, std::size_t direction,
```
tammojan, line 311: This syntax is new to me, what happens here?

DDECal/DDECal.cc
```diff
@@ -978,14 +970,15 @@ bool DDECal::process(const DPBuffer& bufin) {
   return false;
 }
 
-void DDECal::idgCallback(size_t row, size_t direction, size_t dataDescId,
+void DDECal::idgCallback(size_t row, size_t direction,
```
tammojan, line 973: Did you remove this because we only support one datadescid (which is typically a spectral window)? In the future we may want to support more, maybe leave it in here and add an assert that it's zero.  That'll help future programmers (me).

IDGPredict/FacetPredict.h
```diff
@@ -20,384 +20,82 @@
 #define FACET_PREDICT_H
 
 #ifdef HAVE_IDG
-
-#include "DS9FacetFile.h"
```
tammojan, line 24: Where did this go?

tammojan (changes requested):
Overall good changes. However I think we need to show that this actually runs before merging it. I think this bit is currently untested, so we should make a test for it.

Common/buffered_lane.h
mnijhuis-tos: The unit test coverage report says lines 104-106 (in the read() function) are not covered. Although I didn't touch those lines, it would be nice to test if that read() function still works, when lane_->read(...) doesn't return 0. Since that read() function is only used in Interpolate.cc, we'd have to extend tInterpolate.cc (which only has 1 test now) or add a new unit test for buffered_lane.h.

Regarding the repository: This file is in DP3/Common and not in the aocommon subrepo.

DDECal/DDECal.cc
mnijhuis-tos: I'm creating a lambda function that captures 'this'. When the facetpredictor calls it, the lambda forwards all arguments to idgCallback. It's somewhat equivalent to std::bind(idgCallback, this), but written using a more modern syntax.

DDECal/DDECal.cc
mnijhuis-tos: That's indeed the reason why I removed it. I'll store the FacetPredict::PredictCallback, so it includes the dataDescId again, and I'll also update the lambda and idgCallback.

IDGPredict/FacetPredict.h
mnijhuis-tos: Into the new FacetPredict.cc. FacetPredict.h doesn't need this include anymore.

aroffringa:
I briefly looked through the changes and I too think they look good! Nice improvement.

The lane buffer classes are also used in wsclean, so it would be good to put them in aocommon (as long as they are header only -- I think we should keep aocommon header only).

Also I wouldn't proactively change unique_ptr<P> p; p.reset(new ...) to p = boost::make_unique yet, just to avoid too many edits that aren't improving the code, as well as to avoid dependencies on Boost from many files.

IDGPredict/FacetPredict.cc
```diff
@@ -0,0 +1,354 @@
...
+  }
+
+  // TODO, when FacetPredict is a DPStep:
+  // Remove info_ member, and call DPStep::updateInfo(info); here.
+  info_ = info;
+
+  double max_baseline_x = 1.0 / pixel_size_x_;
+  double max_baseline_y = 1.0 / pixel_size_y_;
+  double max_baseline_z = 1.0 / std::min(pixel_size_x_, pixel_size_y_);
+  max_baseline_ = std::sqrt(max_baseline_x * max_baseline_x +
```
tammojan, line 99: Where did this come from?

IDGPredict/FacetPredict.cc
```diff
@@ -0,0 +1,354 @@
...
+    size_t localRowId = rowId - meta.row_id_offset;
+    if (meta.uvws.size() <= localRowId * 3)
+      meta.uvws.resize((localRowId + 1) * 3);
+    for (size_t i = 0; i != 3; ++i) meta.uvws[localRowId * 3 + i] = uvw[i];
+    // IDG uses a flipped coordinate system
+    double uvwFlipped[3] = {uvw[0], -uvw[1], -uvw[2]};
+    while (bs.get_degridder(kDataDescId)
+               ->request_visibilities(rowId, timeIndex, antenna1, antenna2,
+                                      uvwFlipped)) {
+      computePredictionBuffer(direction);
```
tammojan, line 218: Should there be a dataDescId here? (There was at some point.)

IDGPredict/FacetPredict.cc
```diff
@@ -0,0 +1,354 @@
...
+        const double lambda = wavelength(ch);
+        const double u = uvw[0] / lambda;
+        const double v = uvw[1] / lambda;
+        const double w = uvw[2] / lambda;
+        const double angle = u * dlFact + v * dmFact + w * dpFact;
+        const float rotSin = sin(angle);
+        const float rotCos = cos(angle);
+
+        assert(info_.ncorr() == 4);
+        value[0] = {value[0].real() * rotCos - value[0].imag() * rotSin,
```
tammojan, line 266: Why did you remove the loop? Performance?

IDGPredict/FacetPredict.cc
```diff
@@ -0,0 +1,354 @@
...
+      for (size_t ch = 0; ch != info_.nchan(); ++ch) {
+        const double lambda = wavelength(ch);
+        const double u = uvw[0] / lambda;
+        const double v = uvw[1] / lambda;
+        const double w = uvw[2] / lambda;
+        const double angle = u * dlFact + v * dmFact + w * dpFact;
+        const float rotSin = sin(angle);
+        const float rotCos = cos(angle);
+
+        assert(info_.ncorr() == 4);
```
tammojan, line 265: Please move this assert out of the inner loop.

PR #305, opened 2020-09-15
Add tIDGPredict

mnijhuis-tos:
Add integration test that compares wsclean -predict to DDECal with onlypredict


DDECal/test/integration/foursources.reg
```diff
@@ -0,0 +1,21 @@
+# Region file format: DS9 version 4.1
+global color=green dashlist=8 3 width=1 font="helvetica 10 normal roman" select=1 highlite=1 dash=0 fixed=0 edit=1 move=1 delete=1 include=1 source=1
+fk5
+
+# The polygons below were created manually. They create areas around each of the image points.
+# The comments below contain the corresponding 0-based pixel coordinates.
+
+# Ra+dec@(800,128). Polygon is (726, -1) (720, 298) (1061, 313) (1078, 14).
+polygon(22.0,28.0, 22.0,31.0, 18.0,31.0, 18.0,28.0)
```
tammojan, line 9: Depending on which formats we want to support, maybe good to stick to the example format that André gave, i.e. with the text label. We'll need that later on.

IDGPredict/FacetPredict.h
```diff
@@ -91,7 +91,12 @@ class FacetPredict {
   }
 
   void SetMSInfo(std::vector<std::vector<double>>&& bands, size_t nr_stations) {
-    _maxBaseline = 1.0 / std::min(_pixelSizeX, _pixelSizeY);
+    double max_baseline_x = 1.0 / _pixelSizeX;
+    double max_baseline_y = 1.0 / _pixelSizeY;
+    double max_baseline_z = 1.0 / std::min(_pixelSizeX, _pixelSizeY);
+    _maxBaseline = std::sqrt(max_baseline_x * max_baseline_x +
+                             max_baseline_y * max_baseline_y +
```
tammojan, line 98: How is this part of adding a test? Should this be here?

IDGPredict/FacetPredict.h
```diff
@@ -200,7 +206,7 @@ class FacetPredict {
     double uvwr2 = uvw[0] * uvw[0] + uvw[1] * uvw[1] + uvw[2] * uvw[2];
     if (uvw[2] > _maxW && uvwr2 <= _maxBaseline * _maxBaseline) {
       Flush();
-      _maxW *= 1.5;
+      _maxW = std::max(uvw[2], _maxW * 1.5);
```
tammojan, line 209: Same question here. Why did this change?

tammojan:
How are the commits labeled 'improve coverate' and 'Build DPPP in unittest (do not merge)' part of this PR? Did I push them to this branch by accident?

IDGPredict/FacetPredict.h
mnijhuis-tos: It is a fix: Without it, idg ignores some baselines while wsclean does not ignore them. Instead of looking at a single dimension for the pixel size, I'm now considering the size of the pixel 'box' in three dimensions for determining _maxBaseline.

IDGPredict/FacetPredict.h
mnijhuis-tos: The idea is that we are increasing _maxW if it is smaller than uvw[2]. When multiplying _maxW by 1.5 it can still be smaller than uvw[2]. If that's the case, we are increasing _maxW to uvw[2] now.

IDGPredict/FacetPredict.h
tammojan: But where did this code from? Did you make it up yourself, or borrow it from WSClean? If the latter, where?

mnijhuis-tos: I made it up myself.

tammojan (approved):

PR #307, opened 2020-09-16
Add review for DP3 structure

lkrombeenc:


docs/doxygen/structure_review.md
```diff
@@ -0,0 +1,42 @@
...
+
+![Detailed process of a DDECal pipeline](docs/doxygen/images/ddecal.png)
+
+## Identified problems
+We have identified the following impediments in the structure of DP3 that prevent the DDECal step to enable multiple predict types.
+1. A DPStep can only have one input and one output. For example, DDECal requires inputs from three previous step, but can only receive a Buffer from one previous step. This is currently done by having the entire data stream implemented and orchestrated in the DDECal step (except the MSReader and MSWriter).
+2. A DPStep requires a Buffer as input, but it does not always need the (entire) Buffer. For example, the IDGPredict step only needs the UVW and not the entire buffer. Additionally, the second ApplyCal step needs an hdf5 file which cannot be given, but since all processing is done in the DDECal step, the hdf5 file that is needed in that step is already in available.
+
+## Improvements
+1. To solve problem 1, we need to implement internal streams. The process function will no longer have a DPBuffer as input but rather a vector of DPBuffer. An entity orchestrates the process and makes sure that the data flows as described.
```
mnijhuis-tos, line 15:
```suggestion
1. To solve problem 1, the process function should no longer have a single DPBuffer as input but rather a vector of DPBuffers. DP3 should orchestrate the process and make sure that the data flows as described.
```

docs/doxygen/structure_review.md
```diff
@@ -0,0 +1,42 @@
...
+2. DPBuffer will be split into multiple types. For example, UVW will be a buffer on its own. Using a vector as input for the process function, the DPSteps have access to the information they need and nothing more or less. It allows any kind of data type to be streamed in DP3.
```
mnijhuis-tos, line 16:
```suggestion
2. For solving problem 2, we could introduce different DPBuffer types. For example, a buffer then only holds UVW data and nothing else. Using a vector as input for the process function, the DPSteps have access to the information they need and nothing more or less. DP3 then supports streaming any kind of data type.
```

docs/doxygen/structure_review.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Implementation plan
+During the review sessions we concluded that to enable multiple predict types in DDECal not much needs to be done. Therefore, we have decided to not refactor the DP3 structure significantly because we can implement the required feature without changing the core structure of DP3.
+
+We have identified 2 key refactors that fulfil the required use cases:
```
mnijhuis-tos, line 21: "2" -> "two". (For numbers less than 10, using words is more common)

docs/doxygen/structure_review.md
```diff
@@ -0,0 +1,42 @@
...
+1. Refactor the creation of ApplyCal steps in Prediction steps. DDECal should be responsible for the creation of ApplyCal, not Predict. This should therefore be done in the DDECal step and note in the Predict step.
```
mnijhuis-tos, line 22:
```suggestion
1. Refactor the creation of ApplyCal steps in Predict steps. Currently, the DDECal step creates Predict substeps which in turn create ApplyCal subsubsteps. DDECal should be responsible for the creation of ApplyCal, not Predict, so the ApplyCal steps become substeps of DDECal instead of being subsubsteps inside Predict.
```

docs/doxygen/structure_review.md
```diff
@@ -0,0 +1,42 @@
...
+2. Factor out IDGPredict as separate step so that it can be used outside of DDECal.
```
mnijhuis-tos, line 23:
```suggestion
2. Currently, IDGPredict is implemented using a custom Interface. We should factor out this code and convert it to the generic DPStep interface, so it can also be used outside of DDECal.
```

docs/doxygen/structure_review.md
```diff
@@ -0,0 +1,42 @@
...
+3. Refactor the nested DDECal model pipeline so that it can be defined separately for every direction instead of one to rule them all. The metadata must not change in any of the nested model steps of DDECal. This requires a refactors in the process function of DDECal.cc.
```
mnijhuis-tos, line 24:
```suggestion
3. Refactor the nested DDECal model pipeline so that it can be defined separately for every direction instead of one to rule them all. The metadata must not change in any of the nested model steps of DDECal. This requires refactoring the process function of DDECal.cc.
```

docs/doxygen/structure_review.md
```diff
@@ -0,0 +1,42 @@
...
+
+From a user perspective we propose the following additional parset definitions for DDECal:
+- `ddecal.modelnextsteps`: Defines the default steps for the models that need to be applied to every direction before applying DDECal. Each step can have its own parset entries.
```
mnijhuis-tos, line 27:
```suggestion
- `ddecal.modelnextsteps`: Defines the default steps for the models that need to be applied to every direction before using the output for calibration. Each step can have its own parset entries.
```

docs/doxygen/structure_review.md
```diff
@@ -0,0 +1,42 @@
...
+An example of a parset for DDECal using different models for different directions is as follows:
+```
+ddecal.images = [image1.fits, image2.fits]  
+ddecal.regions = ds9.reg  
+ddecal.sourcedb = ATeam.sourcedb  
+ddecal.directions = [CasA, CygA] # subselection of patches from sourcedb
+
+ddecal.modelnextsteps = [applycal3]  
+ddecal.modelnextsteps.CasA = [applycal4, uvwflag3, applybeam3]  
+ddecal.modelnextsteps.CygA = [applycal5, applybeam4]
```
mnijhuis-tos, line 39: At first glance there are only two directions: CasA and CygA, so it looks like applycal3 is not used here. In reality, the regions file also defines directions, so applycal3 will operate on those directions.

Maybe ddecal.directions should include all directions, including those from the regions file. Perhaps we could use special keys like *allregions* or *allpatches* for saying that we want all directions for a given type or input. Alternatively, we could rename ddecal.directions to ddecal.sourcedb.directions so it's clear that it only applies to the patches from sourcedb.

docs/doxygen/structure_review.md
```diff
@@ -0,0 +1,42 @@
...
+ddecal.images = [image1.fits, image2.fits]  
+ddecal.regions = ds9.reg  
+ddecal.sourcedb = ATeam.sourcedb  
+ddecal.directions = [CasA, CygA] # subselection of patches from sourcedb
+
+ddecal.modelnextsteps = [applycal3]  
+ddecal.modelnextsteps.CasA = [applycal4, uvwflag3, applybeam3]  
+ddecal.modelnextsteps.CygA = [applycal5, applybeam4]  
+
+applycal5.h5parm = known_corruptions.h5 # Will automagically get right direction
```
mnijhuis-tos, line 41: I don't understand this line. ddecal.directions already specify the direction, so how does known_corruptions.h5 also automagically set/use/have the right direction?

mnijhuis-tos (changes requested):

docs/doxygen/structure_review.md
```diff
@@ -1,42 +1,54 @@
...
-2. DPBuffer will be split into multiple types. For example, UVW will be a buffer on its own. Using a vector as input for the process function, the DPSteps have access to the information they need and nothing more or less. It allows any kind of data type to be streamed in DP3.
+1. To solve problem 1, a possible solution is that the process function should no longer have a single DPBuffer as input but rather a vector of DPBuffers. DP3 should orchestrate the process and make sure that the data flows as described.
+2. For solving problem 2, we could introduce different DPBuffer types. For example, a buffer then only holds UVW data and nothing else, or a buffer for solutions. Using a vector as input for the process function, the DPSteps have access to the information they need and nothing more or less. DP3 then supports streaming any kind of data type.
+3. Problem 3 can be solved by providing the user with a syntax to define with model is to be applied per direction. DDECal will then initialize the models and applies it to said directions.
```
mnijhuis-tos, line 21:
```suggestion
3. Problem 3 can be solved by providing the user with a syntax to define which model is to be applied per direction. DDECal will then initialize the models and applies it to said directions.
```

docs/doxygen/structure_review.md
```diff
@@ -1,42 +1,54 @@
...
-3. Refactor the nested DDECal model pipeline so that it can be defined separately for every direction instead of one to rule them all. The metadata must not change in any of the nested model steps of DDECal. This requires a refactors in the process function of DDECal.cc.
+We have identified four key refactors that fulfil the required use cases:
+1. Refactor the creation of ApplyCal steps inside Predict steps. Currently, the DDECal step creates Predict substeps which in turn create ApplyCal subsubsteps. DDECal should be responsible for the creation of ApplyCal, not Predict, so the ApplyCal steps become substeps of DDECal instead of being subsubsteps inside Predict.
+2. Currently, IDGPredict is implemented using a custom interface. We should factor out this code and convert it to the generic DPStep interface, so it can also be used outside of DDECal.
+3. Refactor the nested DDECal model pipeline so that it can be defined separately for every direction instead of one to rule them all. The metadata must not change in any of the nested model steps of DDECal. This requires refactoring the steps setup in the constructor of DDECal.cc.
```
mnijhuis-tos, line 31:
```suggestion
3. Refactor the nested DDECal model pipeline so that it can be defined separately for every direction instead of one to rule them all. The metadata must not change in any of the nested model steps of DDECal. This requires refactoring the steps setup in the constructor of DDECal.
```

docs/doxygen/structure_review.md
```diff
@@ -1,42 +1,54 @@
...
+4. Remove the model column from the buffer and create a ColumnReader that can read model columns that can be used in DDECal.
```
mnijhuis-tos, line 32:
```suggestion
4. Remove the model column from the buffer and create a generic ColumnReader that can read any input column, including model columns that can be used in DDECal.
```

docs/doxygen/structure_review.md
```diff
@@ -1,42 +1,54 @@
...
-ddecal.modelnextsteps = [applycal3]  
-ddecal.modelnextsteps.CasA = [applycal4, uvwflag3, applybeam3]  
-ddecal.modelnextsteps.CygA = [applycal5, applybeam4]  
+ddecal.modelnextsteps = [applycal]
+ddecal.modelnextsteps.CasA = [applycal, uvwflag, applybeam]  # Putting a UVWFlag here means applying a UV cut for this direction
+ddecal.modelnextsteps.CygA = [applycal, applybeam]
+ddecal.modelnextsteps.MODEL_DATA_1 = []
 
-applycal5.h5parm = known_corruptions.h5 # Will automagically get right direction
+applycal.h5parm = known_corruptions.h5 # Will automagically get right direction, i.e. the model data in direction CasA will be corrupted with the solution in direction 'CasA'
```
mnijhuis-tos, line 53: corrupted? That doesn't sound good... Maybe say "The solution in direction x will be applied to the model data that has the same direction."

docs/doxygen/structure_review.md
```diff
@@ -1,42 +1,54 @@
...
+4. Remove the model column from the buffer and create a ColumnReader that can read model columns that can be used in DDECal.
 
+## Suggested new parset for DDECal
 From a user perspective we propose the following additional parset definitions for DDECal:
-- `ddecal.modelnextsteps`: Defines the default steps for the models that need to be applied to every direction before applying DDECal. Each step can have its own parset entries.
-- `ddecal.modelnextsteps.<direction>`: Allows steps to be defined for a specific direction. Overrides the default steps for this direction.
+- `ddecal.modelnextsteps`: Defines the default steps for the models that need to be applied to every direction before using the output for calibration. Each step can have its own parset entries.
+- `ddecal.modelnextsteps.<direction>`: Allows steps to be defined for a specific direction. Overrides the default steps for this direction. `<direction>` is the direction name, which is the patch name in the `ddecal.sourcedb`, the region label in the `ddecal.regions`, or the model column name.
```
mnijhuis-tos, line 37:
```suggestion
- `ddecal.modelnextsteps.<direction>`: Allows steps to be defined for a specific direction. Overrides the default steps for this direction. `<direction>` is the direction name, which is the patch name in the `ddecal.sourcedb` database, the region label in the `ddecal.regions` file, or a model column name from `ddecal.modeldatacolumns`.
```

mnijhuis-tos (changes requested):

docs/doxygen/structure_review.md
tammojan: It's jargon, probably users would understand. 'Correct' means 'correct for the instrumental error', whereas 'corrupt' means 'apply the instrumental error to data'.

PR #319, opened 2021-01-13
Update Version.h

revoltek:
Update to version 5


tammojan:
Thanks. I [fixed](https://github.com/lofar-astron/DP3/pull/319) the version to automatically match the cmake version, so we won't forget it next time. So changing manually should not be necessary anymore.

